PR #3, opened 2023-03-01
test(STONEO11Y-43): add promql unit testing

yftacherzog:
Install promtool to allow promql unit testing and provide example test cases.

### DIFF
--- a/automation/test.sh
+++ b/automation/test.sh
@@ -2,7 +2,17 @@
 
 main() {
     ls -l
+    install_promtool
+
     echo "Running tests"
+    ./promtool test rules test/promql/*
+}
+
+install_promtool() {
+    echo "Installing promtool..."
+    VERSION="2.42.0"
+    curl -OJL "https://github.com/prometheus/prometheus/releases/download/v${VERSION}/prometheus-$VERSION.linux-amd64.tar.gz"
+    tar xvzf "prometheus-$VERSION.linux-amd64.tar.gz" "prometheus-$VERSION.linux-amd64"/promtool --strip-components=1
 }
 
 if [[ "${BASH_SOURCE[0]}" == "$0" ]]; then
--- a/test/promql/example.yaml
+++ b/test/promql/example.yaml
@@ -1,0 +1,31 @@
+# This is the main input for unit testing.
+# Only this file is passed as command line argument.
+
+evaluation_interval: 1m
+
+tests:
+    # Test 1.
+    - interval: 1m
+      # Series data.
+      input_series:
+          - series: 'up{job="prometheus", instance="localhost:9090"}'
+            values: '0 0 0 0 0 0 0 0 0 0 0 0 0 0 0'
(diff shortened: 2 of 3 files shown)


Omeramsc:
What/Should there be a difference between example 1 and example 2?

yftacherzog:
@avi-biton autocomplete doesn't work on your name for me. I cannot add you to reviewers (at least for now).

automation/test.sh
```diff
@@ -2,7 +2,13 @@
 
 main() {
     ls -l
+    echo "Installing promtool..."
+    VERSION="2.42.0"
```
eisraeli, line 6: Shouldn't we fetch the latest version [1] instead of having a fixed version.
Otherwise this will have to be updated manually. 

[1] https://github.com/prometheus/prometheus/releases/latest

yftacherzog: The original script I used as source had it as latest. I thought we might prefer having it fixed to avoid surprises.

eisraeli: @yftacherzog Having it as a fixed version often makes more sense if people remember to bump the version manually. From past experience (for example with the Pipfile we had with some of the dependencies having fixed versions), people don't bump the version and then we end up with a very outdated version.
Anyway, I'm fine with both approaches (personally, I prefer yours better to avoid unexpected regressions).

yftacherzog:
> Why do we need 2 identical example files?

https://github.com/redhat-appstudio/o11y/pull/3#issuecomment-1449743197

Omeramsc (approved):
LGTM

automation/test.sh
```diff
@@ -2,7 +2,17 @@
...
+    install_promtool
+
     echo "Running tests"
+    ./promtool test rules test/promql/*
+}
+
+install_promtool() {
+    echo "Installing promtool..."
+    VERSION="2.42.0"
+    curl -OJL "https://github.com/prometheus/prometheus/releases/download/v${VERSION}/prometheus-$VERSION.linux-amd64.tar.gz"
```
ifireball, line 14: Do we really have to curl something for each and every CI run? Can we arrange for this to be preinstalled in whatever image we are using to run this?

Frankly I'm not sure why is there a shell script here at all, shouldn't we be doing things via Tekton tasks?

gbenhaim: this script is being called as part of a tekton task. 
going forward we will probably have an image with all dependencies, but for now, I think this PR is good enough.

ifireball: why run a script rather then write the thing we need to run directly in the task code? what do we gain from adding a level of indirection here?

gbenhaim: ide support, shellcheck, having the ability to run it locally without tekton.

automation/test.sh
amisstea: I think it's fine either way for now. Personally, I also tend to prefer fixed versions for the same reason @eisraeli stated.

amisstea (approved):

PR #4, opened 2023-03-02
chore(STONEO11Y-28): dummy Prometheus recordingRule resource

Omeramsc:
Adding dummy Prometheus recording rules to verify
o11y repository's infrastructure.

The expected behavior is that infra-deployments' ArgoCD instance should automatically deploy any PrometheusRules when there are new commits to the main branch of the 011y repository.

### DIFF
--- a/prometheus/base/kustomization.yaml
+++ b/prometheus/base/kustomization.yaml
@@ -1,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- prometheus.rules.yaml
+- namespaces.yaml
--- a/prometheus/base/namespaces.yaml
+++ b/prometheus/base/namespaces.yaml
@@ -1,0 +1,5 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: o11y
+spec: {}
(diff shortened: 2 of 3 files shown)


yftacherzog (approved):

prometheus/base/kustomization.yaml
```diff
@@ -0,0 +1,2 @@
+resources:
```
gbenhaim, line 1: please add to the top of the file:

```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
```

In addition, you need to define the namespace for the resources.

```
namespace: o11y
```

and add another resource for the namespace definition

```
apiVersion: v1
kind: Namespace
metadata:
  name: o11y
spec: {}
```

gbenhaim:
@Omeramsc in addition to what I wrote above, in the kustomize file in infra-deployments you would need to add
```
commonLabels:
  openshift.io/workload-monitoring: "true"
```
so Prometheus will pick our roles.

prometheus/base/kustomization.yaml
Omeramsc: Thanks! updated

gbenhaim (approved):

PR #5, opened 2023-03-02
test(STONEO11Y-48): add gitlint

yftacherzog:
add gitlint configs and introduce a tekton task to run it.
This will validate the commits in the PR against the rules defined
on the gitlint file.

### DIFF
--- a/.tekton/pull-request.yaml
+++ b/.tekton/pull-request.yaml
@@ -12,11 +12,15 @@ spec:
       value: "{{repo_url}}"
     - name: revision
       value: "{{revision}}"
+    - name: target-branch
+      value: "{{target_branch}}"
   pipelineSpec:
     params:
       - name: git-url
         type: string
       - name: revision
+        type: string
+      - name: target-branch
         type: string
     workspaces:
       - name: workspace
@@ -46,6 +50,30 @@ spec:
             - name: run-tests
               image: registry.access.redhat.com/ubi8/ubi:latest
               script: cd $(workspaces.workspace.path) && ./automation/test.sh
+      - name: run-gitlint
+        params:
+          - name: target-branch
+            value: $(params.target-branch)
+        workspaces:
+          - name: workspace
+            workspace: workspace
+        runAfter:
+          - clone-repository
+        taskSpec:
+          params:
+            - name: target-branch
+          workspaces:
+            - name: workspace
+          steps:
+            - name: run-gitlint
+              image: registry.access.redhat.com/ubi9/python-39:latest
+              workingDir: $(workspaces.workspace.path)
+              script: |
+                #!/bin/bash -ex
+
+                python -m pip install gitlint
+                git fetch --unshallow
+                gitlint --commits "origin/$(params.target-branch)..HEAD"
   workspaces:
     - name: workspace
       volumeClaimTemplate:
--- a/.gitlint
+++ b/.gitlint
@@ -1,0 +1,10 @@
+[general]
+contrib=contrib-title-conventional-commits,contrib-body-requires-signed-off-by
+
+[title-max-length]
+line-length=72
+
+[body-max-line-length]
+line-length=72
+
+[body-first-line-empty]


srivickynesh (approved):
LGTM

yftacherzog:
> LGTM

@srivickynesh it's not triggering at the moment. Trying to figure out why.

srivickynesh:
This will be triggered once it gets merged. you will see it in the next PR.

amisstea:
@yftacherzog is there any reason why we can't use Tekton for this check rather than GitHub actions?

yftacherzog:
> @yftacherzog is there any reason why we can't use Tekton for this check rather than GitHub actions?

I was trying to decide between those, and when searching for examples from other repos on our org, I saw a couple of examples using github actions, but I suppose tekton should be an option as well.

amisstea:
> Do you know of an image on our registry that we can use instead? The one we use on our existing action doesn't have python installed. It doesn't have git either.

You can use [registry.access.redhat.com/ubi9/python-39](https://catalog.redhat.com/software/containers/ubi9/python-39/61a61032bfd4a5234d59629e). It includes git, python and pip preinstalled with no rate limiting.

.tekton/pull-request.yaml
```diff
@@ -46,6 +46,22 @@ spec:
...
+        taskSpec:
+          workspaces:
+            - name: workspace
+          steps:
+            - name: run-gitlint
+              image: registry.access.redhat.com/ubi9/python-39:latest
+              script: |
+                python -m pip install gitlint
+                cd $(workspaces.workspace.path)
+                gitlint --commits $(git merge-base HEAD origin/main)..HEAD
```
yftacherzog, line 64: @amisstea I'm trying here with tekton, as an alternative to github actions. The problem is that it should fail but it's passing and I'm a bit limited with debugging here.

The main issue is that I don't have the logs on one hand, and on the other hand, I don't know which branches are being pulled using the previous task, so it's hard to replicate locally. 

Any suggestions?

amisstea: I tested this out on my fork and discovered an error from the tekton admission controller because the `params` aren't defined in this embedded task.

```shell
{"severity":"ERROR","timestamp":"2023-03-03T17:09:28.691089419Z","logger":"pipelinesascode","caller":"events/emit.go:45","message":"PipelineRun on-pull-request- has failed: creating pipelinerun on-pull-request- in amisstea has failed: admission webhook \"validation.webhook.pipeline.tekton.dev\" denied the request: validation failed: non-existent variable in \"python -m pip install gitlint\\ncd $(workspaces.workspace.path)\\ngitlint --commits $(git merge-base HEAD \\\"$(params.git-url)/$(params.target-branch)\\\")..HEAD\\n\": spec.pipelineSpec.tasks[2].taskSpec.steps[0].script ","commit":"fdfd333","provider":"github","event-id":"273ae060-b9e6-11ed-8b5f-fc68c01af0fc","event-sha":"a5c0d170e192396b804c126550e373b941db787a","event-type":"pull_request","stacktrace":"github.com/openshift-pipelines/pipelines-as-code/pkg/events.(*EventEmitter).EmitMessage\n\t/src/pkg/events/emit.go:45\ngithub.com/openshift-pipelines/pipelines-as-code/pkg/pipelineascode.(*PacRun).Run.func1\n\t/src/pkg/pipelineascode/pipelineascode.g...
```

They need to be declared in both the inline `Task` as well as the underlying `taskRef`. Here's an [example](https://github.com/amisstea/o11y/blob/218c26b451f17c4aa7c935120cc3aa38fe69e6fc/.tekton/pull-request.yaml#L64).

To debug this I used the `tkn-pac` CLI to setup a personal GitHub app which triggers PaC on my cluster. I also had to create a `Repository` CR in the desired namespace. This can all be accomplished using `tkn-pac`.

```shell
# Execute these commands from your fork
tkn-pac bootstrap # Follow the prompts
tkn-pac create repository # Decline the webhook setup during this step
```
A couple things to note:
1. I'm using a hypershift cluster so there was no need to [setup a proxy](https://pipelinesascode.com/docs/install/installation/#proxy-service-for-pac-controller) to get the PaC webhook working.
2. My example above doesn't perform `gitlint` on all the new commits in a pull request, just the head commit. That needs more consideration.
3. Be careful with the embedded subshell you have here (`git merge-base`) because errors from it won't bubble up to the main process. You can work around this by evaluating it on a separate line.

amisstea: @gbenhaim do you know of anything we can do to make debugging CI issues like this easier with the shared infra? It seems like PaC doesn't produce meaningful errors (or anything for that matter) when there are syntax errors in a `PipelineRun`. Also, I didn't have access to the cluster so I'm assuming the rest of the team will run into this limitation as well.

yftacherzog: @amisstea 
> ```shell
> # Execute these commands from your fork
> tkn-pac bootstrap # Follow the prompts
> tkn-pac create repository # Decline the webhook setup during this step
> ```

Is that my olly fork? infra-deployments? I tried with both and am getting:
```
→ tkn-pac bootstrap
🏃 Checking if Pipelines as Code is installed.
Error: Unauthorized
```

yftacherzog: Currently the pipeline fails, while I'm expecting it to pass.
This is what I'm doing at the moment:
```
              script: |
                #!/bin/bash -ex

                python -m pip install gitlint
                git remote add upstream "$(params.git-url)"
                git fetch upstream "$(params.target-branch)"
                BASELINE_COMMIT="$(git merge-base HEAD upstream/"$(params.target-branch)")"
                gitlint --commits "$BASELINE_COMMIT"..HEAD
```
In the little logs I'm getting, I'm seeing this:
```
* [new branch]      main       -> upstream/main
++ git merge-base HEAD upstream/main
+ BASELINE_COMMIT=
```
It seems to me as merge-base fails but I cannot tell why.

yftacherzog: Also, I think merge-base is going to be a problem here in case the PR branch is not rebased on the target branch when this is running, but maybe it will just point to an older commit and won't be a problem.

amisstea: I tested a copy of your branch with my setup and managed to get it [working](https://github.com/amisstea/o11y/pull/4/files). The main issue I found is that the `git-clone` task is performing a shallow clone with `depth`=1 by default. I don't see a way to disable that from the task params (I tried `depth`=0) but there's a simple workaround. I've tested this with a couple malformed commit messages and it raised errors for both with no need for `git merge-base`. Here's the gist of it:

```bash
git fetch --unshallow
gitlint --commits "origin/$(params.target-branch)..HEAD"
```

.tekton/pull-request.yaml
```diff
@@ -46,6 +50,32 @@ spec:
             - name: run-tests
               image: registry.access.redhat.com/ubi8/ubi:latest
               script: cd $(workspaces.workspace.path) && ./automation/test.sh
+      - name: run-gitlint
+        params:
+          - name: git-url
+            value: $(params.git-url)
```
amisstea, line 56: This param is unused and can be removed.

amisstea (approved):

.tekton/pull-request.yaml
srivickynesh: +1

yftacherzog: done

srivickynesh (approved):
LGTM

eisraeli (approved):

.tekton/pull-request.yaml
```diff
@@ -46,6 +50,30 @@ spec:
...
+          steps:
+            - name: run-gitlint
+              image: registry.access.redhat.com/ubi9/python-39:latest
+              workingDir: $(workspaces.workspace.path)
+              script: |
+                #!/bin/bash -ex
+
+                python -m pip install gitlint
+                git fetch --unshallow
+                gitlint --commits "origin/$(params.target-branch)..HEAD"
```
eisraeli, line 76: Wouldn't this just run glint on the main branch and not on the changes in the PR?

yftacherzog: It specifics the commits as far as I understand rather than the branch, so it runs on the PR branch in any case as far as I understand, but I think there's another issue here nonetheless:

I think this will require the PR to always be rebased, which is going to be a headache.

Suppose that you had a PR open and in the meantime another PR was merged, then there's no series of commits getting you from the target branch to the tip of your PR branch. I need to check to see what will happen in that case, but I think we need the merge-base stuff to avoid it.

With merge-base, we should be getting an older ancestor as starting point for the check, which should be ok, as the history from that older ancestor  should be compliant with our gitlint anyway. If not, a rebase should address it, but it's still far less rebasing comparing to what's currently here (if I'm right).

eisraeli: @yftacherzog Yesterday, in my PR, I was trying to for go for the merge-base approach.
But I ran into an issue:
After the cloning task, I ended up with a detached HEAD of the commit in my PR.
It's fine if you just want to scan that commit but pint requires that you'll be on branch X (with history) and point to the base branch ('main' for example).
I couldn't use merge-base because the detached commit didn't have any history (although the PR commit comes from a common ancestor).

yftacherzog: @eisraeli I tested a bit more and I think we're good as-is.

Outputting the log of the checked out ref, we do get the proper log of the PR branch. See example here: https://console-openshift-console.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com/k8s/ns/tekton-ci/tekton.dev~v1beta1~PipelineRun/on-pull-request-j7qrc/logs/run-gitlint

Regarding my concern which may require using merge-base, looking at what gitlint does (running with `--debug`), it uses `git rev-list upstream/main..HEAD`  which gives all the commits on HEAD which are not on upstream/main, even if upstream/main progressed and HEAD is not properly rebased.

PR #9, opened 2023-03-09
test(STONEO11Y-56): extract prometheus rules for tests

yftacherzog:
Our Prometheus rules are stored as Kubernetes objects, but promtool which is used for unit tests requires them to be on their own. This change extract the rules to a rule file as part the script that runs the unit tests.


automation/test.sh
```diff
@@ -1,11 +1,14 @@
-#!/bin/bash -e
+#!/bin/bash -ex
 
 main() {
-    ls -l
     install_promtool
+    install_yq
+
+    echo "Extract Prometheus rules"
+    ./yq ".spec" prometheus/base/prometheus.rules.yaml > test/promql/extracted-rules.yaml
```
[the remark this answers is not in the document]
yftacherzog: The way it looks now, I don't think we intend to have more than one rule, but I might be mistaken. I think we can start with this one an evolve.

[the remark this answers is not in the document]
avi-biton: fine by me :)

avi-biton:
LGTM, very useful

eisraeli:
> Our Prometheus rules are stored as Kubernetes objects, but promtool which is used for unit tests requires them to be on their own. This change extract the rules to a rule file as part the script that runs the unit tests.

Same thing with pint.

Looking at promtool's documentation, I see this:
https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/#syntax-checking-rules

I wonder how much it overlaps with pint... we might be fine just with one of them.

yftacherzog:
> LGTM got to say I expected YAML to have a sort of 'include' mechanism. ):

Was thinking the same myself, and then found this thread: https://stackoverflow.com/a/15437697

automation/test.sh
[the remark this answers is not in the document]
eisraeli: @yftacherzog 
As this also relevant to `pint`, I suggest moving it outside of `test.sh`.  WDYT?

automation/test.sh
```diff
@@ -1,18 +1,39 @@
...
```
eisraeli, line 8: @yftacherzog
As this also relevant to pint, I suggest moving it outside of test.sh. WDYT?

prometheus/base/prometheus.rules.yaml
```diff
@@ -11,3 +11,9 @@ spec:
       rules:
         - record: o11y_prometheus_http_requests_total
           expr: sum by (code) (prometheus_http_requests_total)
+        - record: container:network_transmit_bytes:group
```
eisraeli, line 14: Are these changes related to the extraction of the Prometheus rules ?

test/promql/tests/example.yaml
```diff
@@ -0,0 +1,21 @@
+evaluation_interval: 30s
```
eisraeli, line 1: Are these changes related to the extraction of the Prometheus rules ?

automation/test.sh
yftacherzog: I think taking it out of this script means creating dependencies between tasks.  You'd either need to push this into git - which I don't want (I'm adding those yamls to `.gitignore`) or making them run in sequence somehow sharing the same disk, which seems more of a burden than a gain.

The easiest approach would be to have the linter run as part of this script as well, but I'm not sure we want that.

prometheus/base/prometheus.rules.yaml
yftacherzog: Yes. They make use of a rule. It proves that the extraction works.

test/promql/tests/example.yaml
yftacherzog: Yes. We introduce a rest here that points to the extracted rule file. This proves the process works.

automation/test.sh
yftacherzog: I created another PR on top of this one makes this script specific to Prometheus and provide it with multiple options that will allow it to prepare once and run the relevant test on different executions. Have a look.

prometheus/base/prometheus.rules.yaml
amisstea, line 14: The format of this metric name is inconsistent with the example above. We should establish a naming convention according to [best practices](https://prometheus.io/docs/practices/naming/#metric-names). It seems that we should use `appstudio` ([thread](https://redhat-internal.slack.com/archives/C02CTEB3MMF/p1678308775308909)) for the application prefix.

I suggest we remove the example above and use `appstudio_container_network_transmit_bytes_total` here. Wdyt?

yftacherzog: I don't mind. This is just something I had lying around when creating it. I'll change it.

yftacherzog: done

avi-biton (approved):

PR #10, opened 2023-03-09
chore(STONEO11Y-38): add resource minutes rule

yftacherzog:
Add Prometheus rule to get resource consumption for all
containers and init containers.

The consumer can then sum this over time and filter it for the
required time frame to be billed.


Omeramsc:
Don't we want to add this as a rule as well?

Is there a reason for the 4m specific time duration?

yftacherzog:
> Don't we want to add this as a rule as well?
> 
> Is there a reason for the 4m specific time duration?

I'm not sure what's the best way to add it as a rule. I don't want the `[timeframe]` to be part of the rule, as it will make it applicable only for that time frame.

4m is just one example. There will be more once we agree that this is indeed to way forward.

test/promql/tests/dep-resource-min-1m.yaml
```diff
@@ -0,0 +1,61 @@
+# This is the main input for unit testing.
+# Only this file is passed as command line argument.
```
amisstea, line 2: Please replace this boilerplate with a summary of the tests in this file.

test/promql/tests/dep-resource-min-5m.yaml
```diff
@@ -0,0 +1,38 @@
+# This is the main input for unit testing.
+# Only this file is passed as command line argument.
```
amisstea, line 2: More boilerplate to replace.

test/promql/tests/dep-resource-min-1m.yaml
```diff
@@ -0,0 +1,61 @@
...
+
+evaluation_interval: 1m
+
+rule_files:
+  - ../extracted-rules.yaml
+
+tests:
+  # Test 1.
```
amisstea, line 10: Instead of a generic comment like this, please use a description explaining what this test does in human language. If you prefer to put the descriptions above each `expr` block, that's fine too.

prometheus/base/prometheus.rules.yaml
```diff
@@ -15,3 +15,22 @@ spec:
...
+            group_left(label_pipelines_appstudio_openshift_io_type) kube_pod_labels
+        - record: labeled_containers:resource_minutes:2w
+          expr: |
+            last_over_time(appstudio_container_resource_limits_with_additional_labels[2w])
+            * on (namespace, pod, container) group_left ()
+            (sum by (namespace, pod, container) (increase(container_last_seen[2w] )) / 60)
+    - name: resource_gauge
+      interval: 5m
+      rules:
+        - record: appstudio_container_resource_minutes_gauge
```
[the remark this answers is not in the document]
amisstea: Looks pretty good to me :+1:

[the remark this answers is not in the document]
yftacherzog: @hmariset @Omeramsc the rule defined here can be used for resource minutes for both cpu and memory in both pipelines and deployments (might require additional labels for differentiation between the different uses).

I'm off today (not sure about tomorrow). It can be good if you can start coming up with unit tests for the rule - we can do that on a couple of PR and eventually consolidate them and merge only one of them.

avi-biton, line 32:
```suggestion
        - record: appstudio_container_resource_5minutes_gauge
```

yftacherzog: I'm a bit reluctant about including the window size here as consumer should not care about this.

test/promql/tests/resource-min-5m.yaml
```diff
@@ -0,0 +1,134 @@
...
+      - series: 'kube_pod_labels{namespace="ns1", pod="ns1p1"}'
+        values: '1x12'
+
+    promql_expr_test:
+      - expr: |
+          sum_over_time(appstudio_container_resource_minutes_gauge{resource="r1", namespace="ns1"}[12m])
+        eval_time: 12m
+        exp_samples:
+          - labels: '{resource="r1", namespace="ns1", pod="ns1p1", container="ns1p1c1",}'
+            value: 10
```
yftacherzog, line 112: @amisstea having the rule summing every 5m, we lose some granularity. Namely, we're going to have the same results for every time window that contains the same set of measurements.
```
_ _ _ _ | _ _ _ _ | _ _ _ _ <-- recording intervals
[ _ _ _ | _ _ _ _ | _ _ _ ]  <-- consumer sums over wider window
_ _ _ [ | _ _ _ _ | ] _ _ _ <-- consumer sums over smaller window provides same results
```

yftacherzog: Should we make the rule record every 1m?

prometheus/base/prometheus.rules.yaml
avi-biton: In this case, remove the `minutes`  as it has no meaning here, maybe `appstudio_container_resource_gauge`

yftacherzog: It's still resource minutes gauge. To say, the units of the value it records is resource/minute.

avi-biton: OK, I would however add `limits` somewhere,  `appstudio_container_resource_limits_minutes_gauge` since we are using the `kube_pod_container_resource_limits`.
Just not to get confused by the name with `kube_pod_container_resource_requests`

yftacherzog: Good point. I'll add it.

test/promql/tests/resource-min-5m.yaml
yftacherzog: I opened a ticket for this one.

prometheus/base/prometheus.rules.yaml
yftacherzog: On another thought, the rule is meant to represent resource consumption minutes. Using the limit is just how we decided to address it. If tomorrow we change it to be based on some other metric, I think we might want to keep the same name and just change the implementation.

For that reason, I prefer to leave the name as-is.

avi-biton: So lets keep it for now, we can always change it if it would be needed

yftacherzog:
> LGTM, but I would prefer if every test will have a comment regarding what is it testing. while a good test should be self-explanatory, I believe we can all agree prom tests are everything but readable...

@Omeramsc I agree. Can you point me to where you see this is missing?

hmariset:
LGTM

hmariset (approved):

Kousalya1998 (approved):
lgtm

PR #11, opened 2023-03-09
chore(STONEO11Y-21): add network egress metric and tests

avi-biton:
# add network egress metric

This PR is for adding metric in the rule file that will assist us in getting the network egress (transmit)  
of all the containers, and to filter them by namespace or pipeline type

## Adding a rule file

Adding a rule file with a `container_network_transmit_bytes_total` metric and attach the `label_pipelines_appstudio_openshift_io_type` to the result. 

This way we will be able to filter our results both by `namespace` and `label_pipelines_appstudio_openshift_io_type`  
The new metric is called `appstudio_container_network_transmit_bytes_total`

## Query examples:
- To get the total amount of transmitted bytes in the last day, filtering by namespace:
`sum(increase(appstudio_container_network_transmit_bytes_total[1d])) by (namespace)`

- To get the total amount of transmitted bytes in the last day, filtering by pipeline type:
`sum(increase(appstudio_container_network_transmit_bytes_total[1d])) by (label_pipelines_appstudio_openshift_io_type)`

- These can also be used with the `@` operator to the query to query it at a specific timestamp:
`sum(increase(appstudio_container_network_transmit_bytes_total[1d] @ 1678210200)) by (namespace)`

### DIFF
--- a/prometheus/base/prometheus.rules.yaml
+++ b/prometheus/base/prometheus.rules.yaml
@@ -3,15 +3,16 @@ kind: PrometheusRule
 metadata:
   labels:
     openshift.io/prometheus-rule-evaluation-scope: leaf-prometheus
-  name: o11y-example-rule
+  name: o11y-rules
   namespace: o11y
 spec:
   groups:
-    - name: example
-      rules:
-        - record: appstudio_container_network_transmit_bytes_total
-          expr: |
-            container_network_transmit_bytes_total
-            * on (namespace, pod)
-            group_left (label_pipelines_appstudio_openshift_io_type)
-            kube_pod_labels
+  - name: appstudio-metering
+    interval: 30s
+    rules:
+      - record: appstudio_container_network_transmit_bytes_total
+        expr: |
+          container_network_transmit_bytes_total
+          * on (namespace, pod)
+          group_left (label_pipelines_appstudio_openshift_io_type)
+          kube_pod_labels
--- a/test/promql/tests/test_network.yaml
+++ b/test/promql/tests/test_network.yaml
@@ -1,0 +1,121 @@
+rule_files:
+  - ../extracted-rules.yaml
+
+# interval for evaluating the rules in rule_files
+evaluation_interval: 30s
+
+tests:
+  # Interval for the samples of the series
+  - interval: 30s
+    # Series data
+    input_series:
+
+      - series: 'container_network_transmit_bytes_total{namespace="prod", pod="prod-pod"}'
+        values: '0 1 1 1 4 6 _ _ 8 10 _'
+      - series: 'kube_pod_labels{label_pipelines_appstudio_openshift_io_type="release", namespace="prod", pod="prod-pod"}'
+        values: '1 1 1 1 1 1 1 1 1 1 1'
+
+      - series: 'container_network_transmit_bytes_total{namespace="prod", pod="pre-prod-pod"}'
+        values: '0 2 2 2 4 5 5 0 4 4 5'
+      - series: 'kube_pod_labels{label_pipelines_appstudio_openshift_io_type="", namespace="prod", pod="pre-prod-pod"}'
+        values: '1 1 1 1 1 1 1 1 1 1 1'
+
+      - series: 'container_network_transmit_bytes_total{namespace="test", pod="test-pod"}'
+        values: '0 2 2 2 4 5 5 7 8 9 10'
+      - series: 'kube_pod_labels{label_pipelines_appstudio_openshift_io_type="test", namespace="test", pod="test-pod"}'
+        values: '1 1 1 1 1 1 1 1 1 1 1'
+
+      - series: 'container_network_transmit_bytes_total{namespace="another-test", pod="test-pod"}'
+        values: '0 2 2 2 4 5 5 7 8 9 10'
+      - series: 'kube_pod_labels{label_pipelines_appstudio_openshift_io_type="test", namespace="another-test", pod="test-pod"}'
+        values: '1 1 1 1 1 1 1 1 1 1 1'
+
+
+    promql_expr_test:
+      # Test the core metric with namespace="prod"
+      # Since the sample at 3m is `_` (no sample), the value we get is the last captured value (6)
+      - expr: appstudio_container_network_transmit_bytes_total{namespace="prod"}
+        # The time the query is evaluated
+        eval_time: 3m
+        exp_samples:
+        - labels: 'appstudio_container_network_transmit_bytes_total{namespace="prod", pod="prod-pod", label_pipelines_appstudio_openshift_io_type="release"}'
+          value: 6
+        - labels: 'appstudio_container_network_transmit_bytes_total{namespace="prod", pod="pre-prod-pod"}'
+          value: 5
+
+      # Test the core metric with namespace="test"
+      - expr: appstudio_container_network_transmit_bytes_total{namespace="test"}
+        eval_time: 4m
+        exp_samples:
+        - labels: 'appstudio_container_network_transmit_bytes_total{namespace="test", pod="test-pod", label_pipelines_appstudio_openshift_io_type="test"}'
+          value: 8
+
+      # Test the core metric with namespace="another-test"
+      - expr: appstudio_container_network_transmit_bytes_total{namespace="another-test"}
+        eval_time: 4m
+        exp_samples:
+        - labels: 'appstudio_container_network_transmit_bytes_total{namespace="another-test", pod="test-pod", label_pipelines_appstudio_openshift_io_type="test"}'
+          value: 8
+
+      # Test the increase in the last 4 minutes after 5 minutes of samples, filtered by namespace="prod"
+      - expr: increase(appstudio_container_network_transmit_bytes_total{namespace="prod"}[4m])
+        eval_time: 5m
+        exp_samples:
+        # We have 2 results:
+        #    one for the results that includes the label_pipelines_appstudio_openshift_io_type="release"
+        #    and one for label_pipelines_appstudio_openshift_io_type=""
+        #
+        # The increase function range is [5m-4m] to [5m]
+        # Since the sample in 5m is `_` we get the last captured value (10)
+        # so the increase between 1 and 10 is 9
+        - labels: '{namespace="prod", pod="prod-pod", label_pipelines_appstudio_openshift_io_type="release"}'
+          value: 9
+
+        # In the middle of the samples we collected a 0 sample and the count was reset.
+        # In this case, the `increase` function knows to handle resets by comparing
+        # the values to the previous values, if the value is lower than the value before,
+        # the value is added to the previous value and the counter continues.
+        # In our example. we have:
+        # values: '0 2 2 2 4 5 5 0 4 4 5'
+        # When increase detects the reset (0, followed by 4), it adds the (0) to the last count and increase all
+        # the values afterwards so '0 2 2 2 4 5 5 0 4 4 5' becomes '0 2 2 2 4 5 5 5 9 9 10'
+        - labels: '{namespace="prod", pod="pre-prod-pod"}'
+          value: 8
+
+      # Test the sum by namespace of the increase in the last 4 minutes after 5 minutes of samples
+      # Basically that's the total amount of bytes transmitted in the last 4 minutes by namespace
+      - expr: sum(increase(appstudio_container_network_transmit_bytes_total[4m])) by (namespace)
+        eval_time: 5m
+        exp_samples:
+        - labels: '{namespace="prod"}'
+          value: 17
+        - labels: '{namespace="test"}'
+          value: 8
+        - labels: '{namespace="another-test"}'
+          value: 8
+
+      # Test the sum by label_pipelines_appstudio_openshift_io_type of the increase in the last 4 minutes after 5 minutes of samples
+      # Basically that's the total amount of bytes transmitted in the last 4 minutes by label_pipelines_appstudio_openshift_io_type
+      - expr: sum(increase(appstudio_container_network_transmit_bytes_total[4m])) by (label_pipelines_appstudio_openshift_io_type)
+        eval_time: 5m
+        exp_samples:
+        - labels: '{label_pipelines_appstudio_openshift_io_type="test"}'
+          value: 16
+        - labels: '{label_pipelines_appstudio_openshift_io_type="release"}'
+          value: 9
+        - labels: '{}'
+          value: 8
+
+      # Test the total amount of bytes transmitted in the last 4 minutes by label_pipelines_appstudio_openshift_io_type="test"
+      - expr: sum(increase(appstudio_container_network_transmit_bytes_total{label_pipelines_appstudio_openshift_io_type="test"}[4m]))
+        eval_time: 5m
+        exp_samples:
+        - labels: '{}'
+          value: 16
+
+      # Test the total amount of bytes transmitted in the last 4 minutes by label_pipelines_appstudio_openshift_io_type that has value in
+      - expr: sum(increase(appstudio_container_network_transmit_bytes_total{label_pipelines_appstudio_openshift_io_type!=""}[4m]))
+        eval_time: 5m
+        exp_samples:
+        - labels: '{}'
+          value: 25


prometheus/base/prometheus.rules.yaml
```diff
@@ -1,13 +1,12 @@
...
-  name: o11y-example-rule
-  namespace: o11y
-spec:
-  groups:
-    - name: example
-      rules:
-        - record: o11y_prometheus_http_requests_total
-          expr: sum by (code) (prometheus_http_requests_total)
+
+groups:
```
Omeramsc, line 2: Why are we removing k8s settings? I thought @yftacherzog #9 fix that issue for now?
do note we have a patch on Infra Deployments that pull updates for the rules file

prometheus/base/prometheus.rules.yaml
```diff
@@ -1,13 +1,12 @@
...
+  - name: stonesoup-metering
+    interval: 30s
+    rules:
+       - record: container:network_transmit_bytes:group
+         expr: |2
+            container_network_transmit_bytes_total
+            * on (namespace, pod)
```
Omeramsc, line 9: does that include init containers?

prometheus/base/prometheus.rules.yaml
avi-biton: Its like that because #9 is not merged yet, and I wanted you to take a look at the work in progress

prometheus/base/prometheus.rules.yaml
```diff
@@ -1,13 +1,12 @@
...
-      rules:
-        - record: o11y_prometheus_http_requests_total
-          expr: sum by (code) (prometheus_http_requests_total)
+
+groups:
+  - name: stonesoup-metering
+    interval: 30s
+    rules:
+       - record: container:network_transmit_bytes:group
+         expr: |2
```
yftacherzog, line 7:
```suggestion
         expr: |
```
As far as I understand, `|2` means that you should have an indentation of `2` for content of `expr`. In reality, I think you have more. I think you can just drop the number for now.

See some details:
https://stackoverflow.com/questions/15540635/what-is-the-use-of-the-pipe-symbol-in-yaml

test/promql/test_network.yml
```diff
@@ -0,0 +1,108 @@
...
+
+rule_files:
+    - ../../prometheus/base/prometheus.rules.yaml
+
+evaluation_interval: 30s
+
+tests:
+ - interval: 30s
+   # Series data 
+   input_series:
```
yftacherzog, line 12: Looks good!

I think we should introduce inputs with missing data. i.e. `_` readings, and verify the results make sense.

I imagine that the way bugs will manifests for this sort of code is through readings that we did not expect to happen, and our way to recreate them will be generate inputs that behave this way.

We don't need to try and anticipate them all now, but the once we know of, we should already introduce (e.g. recordings for pods that are already dead/deleted).

avi-biton: Yes, I intend to add both `_` and `0` to simulate restart of count.

prometheus/base/prometheus.rules.yaml
avi-biton: I didn't see any metric or query to measure network in init containers.  
I will keep on looking...

prometheus/base/prometheus.rules.yaml
avi-biton: Done :)

prometheus/base/prometheus.rules.yaml
Omeramsc: Seems like it really does not exist. 
If we can expect some of the init_container name (e.g: it starts with `init_`)
we can try filtering with the container name... 
although that's just a filter on the query we already use, so I'm not sure why it does not apply for init_containers.

avi-biton: Is there network activity in init containers?

test/promql/test_network.yml
avi-biton: I updated the tests to include both `0` and `_`.  
I also added some explanations for the tests.

prometheus/base/prometheus.rules.yaml
Omeramsc: AFAIK anything can happen in an init container, so it depends...

eisraeli: If we want to monitor network traffic from init containers as well, shouldn't the `container` label be added to the `on` clause?
For example:
`* on (namespace, pod, container) `

avi-biton: The only container that appears in the "core" metric `container_network_transmit_bytes_total` is the `POD` container, so adding it will not solve this.

I cannot find any init_container related metrics other the the resource limits

Omeramsc: upon checking, I think we can assume for now there is no egress usage from init containers. 
even when the query is not filtered, we cannot get egress info, so filters won't help here, and there is no specific query for the init containers egress. This is something that can be solved (if needed) only with extreme measures we've discussed and set aside for the current solution, such as Ebpf.

avi-biton: I will test it today by creating a pod with container and init container, both will run some network egress, and see if I can get the expected total egress

test/promql/tests/test_network.yaml
```diff
@@ -0,0 +1,123 @@
...
+      - series: 'container_network_transmit_bytes_total{namespace="another-test", pod="test-pod"}'
+        values: '0 2 2 2 4 5 5 7 8 9 10'
+      - series: 'kube_pod_labels{label_pipelines_appstudio_openshift_io_type="test", namespace="another-test", pod="test-pod"}'
+        values: '1 1 1 1 1 1 1 1 1 1 1'
+
+
+    promql_expr_test:
+
+      # Test the core metric with namespace=prod
+      # Since the sample at 3m is `_` (no sample), the value we get is the last captures value (6)
```
Omeramsc, line 38: typo(?): captured

test/promql/tests/test_network.yaml
```diff
@@ -0,0 +1,123 @@
...
+        eval_time: 5m
+        exp_samples:
+        - labels: '{label_pipelines_appstudio_openshift_io_type="test"}'
+          value: 16
+        - labels: '{label_pipelines_appstudio_openshift_io_type="release"}'
+          value: 9
+        - labels: '{}'
+          value: 8
+
+      # Test the total amount of bytes transmited in the last 4 minutes by label_pipelines_appstudio_openshift_io_type=test
```
Omeramsc, line 111: isn't that 5m?

test/promql/tests/test_network.yaml
avi-biton: Yes, I will correct it :)

test/promql/tests/test_network.yaml
avi-biton: The `eval_time` is 5 minutes, it means that our metric runs 5 minutes now, and the last sample is at 5 minutes.
We are evaluating the last 4 minutes (`[4m]`)of the `eval_time` so it is the last 4 minutes out of the total 5

yftacherzog:
can we squash the commits?

Kousalya1998 (approved):
lgtm

prometheus/base/prometheus.rules.yaml
```diff
@@ -7,11 +7,12 @@ metadata:
...
   groups:
-    - name: example
-      rules:
-        - record: appstudio_container_network_transmit_bytes_total
-          expr: |
-            container_network_transmit_bytes_total
-            * on (namespace, pod)
-            group_left (label_pipelines_appstudio_openshift_io_type)
-            kube_pod_labels
+  - name: stonesoup-metering
```
amisstea, line 10:
```suggestion
  - name: appstudio-metering
```

test/promql/tests/test_network.yaml
```diff
@@ -0,0 +1,123 @@
+# This is the main input for unit testing.
+# Only this file is passed as command line argument.
+
+rule_files:
+  - ../extracted-rules.yaml
+
+evaluation_interval: 30s
+
+tests:
+  - interval: 30s
```
Omeramsc, line 10: If we already have readability comments, I'd add a short comment about each value in a series being a result of a 30s interval

test/promql/tests/test_network.yaml
```diff
@@ -0,0 +1,123 @@
...
+      # Since the sample at 3m is `_` (no sample), the value we get is the last captured value (6)
+      - expr: appstudio_container_network_transmit_bytes_total{namespace="prod"}
+        eval_time: 3m
+        exp_samples:
+        - labels: 'appstudio_container_network_transmit_bytes_total{namespace="prod", pod="prod-pod", label_pipelines_appstudio_openshift_io_type="release"}'
+          value: 6
+        - labels: 'appstudio_container_network_transmit_bytes_total{namespace="prod", pod="pre-prod-pod"}'
+          value: 5
+
+      # Test the core metric with namespace=test
```
Omeramsc, line 47: I'll write the namespace name with quotations for easier CTRL+F ('# Test the core metric with "namespace=test"'
(same for all other comments)

test/promql/tests/test_network.yaml
```diff
@@ -0,0 +1,123 @@
...
+      - expr: increase(appstudio_container_network_transmit_bytes_total{namespace="prod"}[4m])
+        eval_time: 5m
+        exp_samples:
+        # We have 2 results:
+        #    one for the results that includes the label_pipelines_appstudio_openshift_io_type="release"
+        #    and one for label_pipelines_appstudio_openshift_io_type=""
+        #
+        # The increase function range is [5m-4m] to [5m]
+        # Since the sample in 5m is `_` we get the last captured value (10)
+        # so the increase between 1 and 10 is 9
```
Omeramsc, line 72: Is the increase calculation made between the base metric and the label metric? (is that where the "1" from?)

test/promql/tests/test_network.yaml
```diff
@@ -0,0 +1,123 @@
...
+      # Basically that's the total amount of bytes transmited in the last 4 minutes by label_pipelines_appstudio_openshift_io_type
+      - expr: sum(increase(appstudio_container_network_transmit_bytes_total[4m])) by (label_pipelines_appstudio_openshift_io_type)
+        eval_time: 5m
+        exp_samples:
+        - labels: '{label_pipelines_appstudio_openshift_io_type="test"}'
+          value: 16
+        - labels: '{label_pipelines_appstudio_openshift_io_type="release"}'
+          value: 9
+        - labels: '{}'
+          value: 8
```
Omeramsc, line 109: I expected 9 here. which delta is calculated to produce 8?
```
      - series: 'container_network_transmit_bytes_total{namespace="prod", pod="pre-prod-pod"}'
        values: '0 2 2 2 4 5 5 0 4 4 5'  // --> 0 2 2 2 4 5 5 5 **9** 9 10
      - series: 'kube_pod_labels{label_pipelines_appstudio_openshift_io_type="", namespace="prod", pod="pre-prod-pod"}'
        ```

Comment (Omeramsc) [reply test/promql/tests/test_network.yaml:109]: or is it taking `10` as value due to the shorter eval_time? (which still is not 8)

File: test/promql/tests/test_network.yaml [hunk 12]
```diff
@@ -0,0 +1,123 @@
...
+
+      # Test the sum by label_pipelines_appstudio_openshift_io_type of the increase in the last 4 minutes after 5 minutes of samples
+      # Basically that's the total amount of bytes transmited in the last 4 minutes by label_pipelines_appstudio_openshift_io_type
+      - expr: sum(increase(appstudio_container_network_transmit_bytes_total[4m])) by (label_pipelines_appstudio_openshift_io_type)
+        eval_time: 5m
+        exp_samples:
+        - labels: '{label_pipelines_appstudio_openshift_io_type="test"}'
+          value: 16
+        - labels: '{label_pipelines_appstudio_openshift_io_type="release"}'
+          value: 9
```
Omeramsc, line 107: also expected 8 here as it is the 9th number on the release. what am I doing wrong? :)

test/promql/tests/test_network.yaml
```diff
@@ -0,0 +1,123 @@
...
+        - labels: '{}'
+          value: 8
+
+      # Test the total amount of bytes transmited in the last 4 minutes by label_pipelines_appstudio_openshift_io_type=test
+      - expr: sum(increase(appstudio_container_network_transmit_bytes_total{label_pipelines_appstudio_openshift_io_type="test"}[4m]))
```
Omeramsc, line 112: isn't the test above does the exact same thing?

Omeramsc (changes requested):
Lest some minor comments and some questions

test/promql/tests/test_network.yaml
avi-biton: Done

prometheus/base/prometheus.rules.yaml
avi-biton: Done

test/promql/tests/test_network.yaml
avi-biton: Done

test/promql/tests/test_network.yaml
avi-biton: No, the increase is done on `appstudio_container_network_transmit_bytes_total` which is the base metric `container_network_transmit_bytes_total` with the added `label_pipelines_appstudio_openshift_io_type` label

test/promql/tests/test_network.yaml
avi-biton: Same explanation as previous

test/promql/tests/test_network.yaml
avi-biton: The test above filters by all pipelines types and display all of them.  
This test is for displaying a specific pipeline type, and only shows it.

avi-biton:
> can we squash

Tried to with Omer and Gal, there is something strange that we could not solve, so no squashing fpr now

test/promql/tests/test_network.yaml
Omeramsc: ack, thanks

prometheus/base/prometheus.rules.yaml
```diff
@@ -7,11 +7,12 @@ metadata:
   namespace: o11y
```
srivickynesh, line 7: this has to be on openshift-user-workload-monitoring namespace. or any namespace that we want to apply this rule file to?

avi-biton: It applies to all namespaces, it should be applied for this namespace as it is the namespace of the user workload Prometheus instance

yftacherzog (changes requested):
Let's have the git stuff figured out before we merge.

prometheus/base/prometheus.rules.yaml
avi-biton: Thanks Alex

test/promql/tests/test_network.yaml
```diff
@@ -0,0 +1,121 @@
...
+          value: 8
+
+      # Test the total amount of bytes transmited in the last 4 minutes by label_pipelines_appstudio_openshift_io_type="test"
+      - expr: sum(increase(appstudio_container_network_transmit_bytes_total{label_pipelines_appstudio_openshift_io_type="test"}[4m]))
+        eval_time: 5m
+        exp_samples:
+        - labels: '{}'
+          value: 16
+
+      # Test the total amount of bytes transmited in the last 4 minutes by label_pipelines_appstudio_openshift_io_type that has value in
```
yftacherzog, line 116: typo: transmitted

test/promql/tests/test_network.yaml
```diff
@@ -0,0 +1,121 @@
...
+        eval_time: 5m
+        exp_samples:
+        - labels: '{label_pipelines_appstudio_openshift_io_type="test"}'
+          value: 16
+        - labels: '{label_pipelines_appstudio_openshift_io_type="release"}'
+          value: 9
+        - labels: '{}'
+          value: 8
+
+      # Test the total amount of bytes transmited in the last 4 minutes by label_pipelines_appstudio_openshift_io_type="test"
```
yftacherzog, line 109: typo: transmitted

test/promql/tests/test_network.yaml
```diff
@@ -0,0 +1,121 @@
...
+        exp_samples:
+        - labels: '{namespace="prod"}'
+          value: 17
+        - labels: '{namespace="test"}'
+          value: 8
+        - labels: '{namespace="another-test"}'
+          value: 8
+
+      # Test the sum by label_pipelines_appstudio_openshift_io_type of the increase in the last 4 minutes after 5 minutes of samples
+      # Basically that's the total amount of bytes transmited in the last 4 minutes by label_pipelines_appstudio_openshift_io_type
```
yftacherzog, line 98: typo: transmitted

test/promql/tests/test_network.yaml
```diff
@@ -0,0 +1,121 @@
...
+        # the value is added to the previous value and the counter continues.
+        # In our example. we have:
+        # values: '0 2 2 2 4 5 5 0 4 4 5'
+        # When increase detects the reset (0, followed by 4), it adds the (0) to the last count and increase all
+        # the values afterwards so '0 2 2 2 4 5 5 0 4 4 5' becomes '0 2 2 2 4 5 5 5 9 9 10'
+        - labels: '{namespace="prod", pod="pre-prod-pod"}'
+          value: 8
+
+      # Test the sum by namespace of the increase in the last 4 minutes after 5 minutes of samples
+      # Basically that's the total amount of bytes transmited in the last 4 minutes by namespace
```
yftacherzog, line 86: typo: transmitted

test/promql/tests/test_network.yaml
```diff
@@ -0,0 +1,121 @@
...
+        #    and one for label_pipelines_appstudio_openshift_io_type=""
+        #
+        # The increase function range is [5m-4m] to [5m]
+        # Since the sample in 5m is `_` we get the last captured value (10)
+        # so the increase between 1 and 10 is 9
+        - labels: '{namespace="prod", pod="prod-pod", label_pipelines_appstudio_openshift_io_type="release"}'
+          value: 9
+
+        # In the middle of the samples we collected a 0 sample and the count was reset.
+        # In this case, the `increase` function knows to handle resests by comparing
```
yftacherzog, line 75: typo: resets

yftacherzog (approved):
Few typos, but other than that lgtm. I think it's a good collection of corner cases.

prometheus/base/prometheus.rules.yaml
yftacherzog, line 1: Can we take this opportunity to also rename the PrometheusRule object? i.e. to change the `name: o11y-example-rule` line.

test/promql/tests/test_network.yaml
avi-biton: Done

test/promql/tests/test_network.yaml
avi-biton: Done

test/promql/tests/test_network.yaml
avi-biton: Done

test/promql/tests/test_network.yaml
avi-biton: Done

test/promql/tests/test_network.yaml
avi-biton: Done

prometheus/base/prometheus.rules.yaml
avi-biton: I changed it to `o11y-rules`

prometheus/base/prometheus.rules.yaml
avi-biton: As we saw in our tests, only the first `init container` (if defined) would be detected, no other init containers or containers would be detected afterwards.
This is coming from `cAdvisor`...

yftacherzog: Let's open a ticket for that so we try to address it later on.

Omeramsc (approved):
LGTM

PR #12, opened 2023-03-13
test(STONEO11Y-49): add yamllint

srivickynesh:
add .yamllint config and yaml-lint to CI.

### DIFF
--- a/.tekton/pull-request.yaml
+++ b/.tekton/pull-request.yaml
@@ -74,6 +74,24 @@ spec:
                 python -m pip install gitlint
                 git fetch --unshallow
                 gitlint --commits "origin/$(params.target-branch)..HEAD"
+      - name: yaml-lint
+        workspaces:
+          - name: workspace
+            workspace: workspace
+        runAfter:
+          - clone-repository
+        taskSpec:
+          workspaces:
+            - name: workspace
+          steps:
+            - name: yaml-lint
+              image: registry.access.redhat.com/ubi9/python-39:latest
+              workingDir: $(workspaces.workspace.path)
+              script: |
+                #!/bin/bash -ex
+
+                python -m pip install yamllint
+                yamllint .
   workspaces:
     - name: workspace
       volumeClaimTemplate:
--- a/.yamllint
+++ b/.yamllint
@@ -1,0 +1,26 @@
+rules:
+  braces: enable
+  brackets: enable
+  colons: enable
+  commas: enable
+  comments:
+    level: warning
+  comments-indentation:
+    level: warning
+  document-end: disable
+  document-start: disable
+  empty-lines: enable
+  empty-values: enable
+  hyphens: enable
+  key-duplicates: enable
+  key-ordering: disable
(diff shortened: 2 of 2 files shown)


.tekton/pull-request.yaml
```diff
@@ -74,6 +74,25 @@ spec:
...
+            - name: workspace
+          steps:
+            - name: yaml-lint
+              image: registry.access.redhat.com/ubi9/python-39:latest
+              workingDir: $(workspaces.workspace.path)
+              script: |
+                #!/bin/bash -ex
+
+                python -m pip install yamllint
+                cd $(workspaces.workspace.path)
```
yftacherzog, line 94: Do we need this line if we already specify `workingDir`?

.tekton/pull-request.yaml
```diff
@@ -74,6 +74,25 @@ spec:
...
+                yamllint .
```
yftacherzog, line 95: Will it check every yaml on every sub directory?

srivickynesh: yes

.tekton/pull-request.yaml
srivickynesh: just checking here to see if that is the root path is correct

.tekton/pull-request.yaml
yftacherzog: ok. The guidelines will be something to be added to the contributing guide.

Also, when it comes to Prometheus unit tests, I didn't find a way yet to break long lines, and they tend to be very long. We need to either find a way to do it, or find a way to exempt them specifically from line length tests.

srivickynesh: yeah. I will also add the rule files for yamllint

yftacherzog: cool

srivickynesh: Removed the long-line errors

.tekton/pull-request.yaml
srivickynesh: The root path is correct and removed it.

yftacherzog:
lgtm

.yamllint
```diff
@@ -0,0 +1,35 @@
...
+  quoted-strings: disable
+  trailing-spaces: enable
+  truthy:
+    level: warning
+
+  # accept both     key:
+  #                   - item
+  #
+  # and             key:
+  #                 - item
```
yftacherzog, line 33: Do we need this?

srivickynesh: No, we don't need it. will remove it.

yftacherzog (approved):

.yamllint
```diff
@@ -0,0 +1,29 @@
+ignore: |
+  /vendor
```
amisstea, line 2: Not a big deal but why is this needed?

amisstea (approved):

.yamllint
srivickynesh: Removed

PR #16, opened 2023-03-16
Added additional test cases for resource limits

hmariset:


test/promql/tests/pipelinerun-cpumins.yaml
```diff
@@ -0,0 +1,136 @@
...
+      - series: 'kube_pod_container_resource_limits{resource="memory", namespace="ns2", pod="ns2p1", container="ns2p1c1"}'
+        values: '1 1 1 4 _ 5 _ 7 9 10 _'
+      - series: 'kube_pod_container_resource_limits{resource="cpu", namespace="ns2", pod="ns2p1", container="ns2p1c1"}'
+        values: '0 5 8 1 _ 10 11 12 12 _ 13'
+      - series: 'kube_pod_labels{label_pipelines_appstudio_openshift_io_type="test", namespace="ns2", pod="ns2p1"}'
+        values: '1x10'
+      - series: 'container_last_seen{namespace="ns2", pod="ns2p1", container="ns2p1c2"}'
+        values: '10 10 20 30 40 0 30 40 60 70 70'
+      - series: 'container_last_seen{namespace="ns2", pod="ns2p1", container="ns2p1c1"}'
+        values: '10 10 20 40 0 40 50 50 60 70 70'
```
yftacherzog, line 39: container_last_seen is a timestamp (unix time in seconds) for: `Last time a container was seen by the exporter` (see here: https://github.com/google/cadvisor/blob/master/docs/storage/prometheus.md#prometheus-container-metrics).

It means that if our measurements interval is 60s, and the container continues to live throughout all measurements, then the value of this metric should grow by 60 on every interval.

It's also possible that it dies somewhere in between measurements, so you may have a single value which is an increase of less than 60 comparing to the previous one, but the measurements afterwards cannot grow with something less than 60.

Let's try to visualize (the actual timestamp is going to look something like `1678956539`, but we're using smaller numbers for convenience):

In this case, c1 is a container that lived throughout the entire measurements, and c2 died just before 60s and somehow resurrected (not sure if that is possible, but anyway) between 180s and 240s.
You see that it has 2 measurements of 57 and afterwards one missing value. I think in reality, it's not going to show the same value twice but instead a value will indeed be missing, but anyhow.

| epoc  | 0 | 60 | 120 | 180 | 240 | 300 | 360 |
| ---  | --- | --- | --- | --- | --- | --- | --- |
| c1_last_seen  | 0 | 60 | 120 | 180 | 240 | 300 | 360 |
| c2_last_seen  | 0 | 57 | 57 | _ | 240 | 300 | 360 |

In other words, any new value appearing will have to be from within the epoc values of the time window it was taken in.

Hope it makes sense...

eisraeli: Very nice explanation. 
But why would be a measurement of 57 of 57?
Wouldn't the series look like this?

0 | 57 | _ | _ | 240 | 300 | 360

yftacherzog: > I think in reality, it's not going to show the same value twice but instead a value will indeed be missing, but anyhow.

test/promql/tests/pipelinerun-cpumins.yaml
```diff
@@ -0,0 +1,127 @@
...
+
+      - series: 'kube_pod_init_container_resource_limits{resource="memory", namespace="ns1", pod="ns1p1", container="ns1p1c2"}'
+        values: '2x10'
+      - series: 'kube_pod_init_container_resource_limits{resource="cpu", namespace="ns1", pod="ns1p1", container="ns1p1c2"}'
+        values: '0.5x10'
+      - series: 'container_last_seen{namespace="ns1", pod="ns1p1", container="ns1p1c2"}'
+        values: '100+60x10'
+
+      - series: 'kube_pod_container_resource_limits{resource="memory", namespace="ns2", pod="ns2p1", container="ns2p1c1"}'
+        values: '1 1 1 4 _ 5 _ 7 9 10 _'
```
avi-biton, line 31: Since we are talking about predefined fixed limits, their value is not suppose to change, it is not a counter.
If my memory limit is `1000` it will remain `1000` for all the container's life.
So a better representation of resource limits would be:
```
- series: 'kube_pod_container_resource_limits{resource="memory", namespace="ns2", pod="ns2p1", container="ns2p1c1"}'
        values: '1000 1000 1000 1000 1000 1000 1000 1000 1000 1000'
```

Or 
```
- series: 'kube_pod_container_resource_limits{resource="memory", namespace="ns2", pod="ns2p1", container="ns2p1c1"}'
        values: '1000x10'
```

test/promql/tests/pipelinerun-cpumins.yaml
```diff
@@ -0,0 +1,127 @@
...
+      - series: 'kube_pod_container_resource_limits{resource="cpu", namespace="ns2", pod="ns2p1", container="ns2p1c1"}'
+        values: '0 5 8 1 _ 10 11 12 12 _ 13'
```
avi-biton, line 33: Same comment as the previous

test/promql/tests/pipelinerun-cpumins.yaml
```diff
@@ -0,0 +1,127 @@
...
+            value: 2
+          - labels: 'appstudio_container_resource_limits_with_additional_labels{resource="cpu", namespace="ns1", pod="ns1p1", container="ns1p1c1", label_pipelines_appstudio_openshift_io_type="t1"}'
+            value: 0.5
+          - labels: 'appstudio_container_resource_limits_with_additional_labels{resource="cpu", namespace="ns1", pod="ns1p1", container="ns1p1c2", label_pipelines_appstudio_openshift_io_type="t1"}'
+            value: 0.5
+          - labels: 'appstudio_container_resource_limits_with_additional_labels{resource="cpu", namespace="ns2", pod="ns2p1", container="ns2p1c1", label_pipelines_appstudio_openshift_io_type="test"}'
+            value: 1
+
+      - expr: |
+          increase(container_last_seen{namespace="ns1"}[4m])
```
yftacherzog, line 61: What's the reasoning behind having this tested?

yftacherzog: Same goes for the other tests involving just metrics provided out of the box.

Omeramsc:
As Yfatch's PR covers all three of our efforts, we may be able to close this PR once it merges.

test/promql/tests/pipelinerun-cpumins.yaml
hmariset: > As Yfatch's PR covers all three of our efforts, we may be able to close this PR once it merges.

Sure, will close this once it merged

PR #17, opened 2023-03-16
chore(STONEO11Y-61): add makefile

yftacherzog:
Run tests using makefile. This will allow developers running test prep steps together with the tests themselves using `make all`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,18 @@
+.PHONY: all
+all: prepare lint test_rules
+
+.PHONY: prepare
+prepare:
+	./automation/promtool_tests.sh prepare
+
+.PHONY: lint
+lint:
+	./automation/promtool_tests.sh lint
+
+.PHONY: test_rules
+test_rules:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -0,0 +1,18 @@
+.PHONY: all
+all: prepare lint test_rules
```
eisraeli, line 2: Don't you want to run `pint` as well ?

yftacherzog: I ran it locally and noticed it reinstall pint every time. It also not on gitignore so having it here increases the risk of it being pushed to git. I opened a ticket to address those. Once addressed, we should add it as well here.

BTW, target `all` is only run in local dev envs. on CI, the targets are run independently according to the ci step, so pint is running there.

avi-biton (approved):

avi-biton:
LGTM

eisraeli (approved):
LGTM

Omeramsc (approved):
LGTM

PR #19, opened 2023-03-20
chore: add owners file

yftacherzog:
Add owners file which should hopefully address issues causing the ci not to run in some cases.


yftacherzog:
@gbenhaim do we need anything else other than this file? is this the right content?

gbenhaim (approved):

PR #21, opened 2023-03-21
chore(STONEO11Y-63): Add pint to gitignore & install in Makefile

eisraeli:
* pint installation have being moved from `pint_lint` to the Makefile. This saves a check if pint is already installed.

* Added pint to .gitignore

* Added pint to the makefile `all` target

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,10 @@
+PINT_VERSION='0.42.2'
+
 .PHONY: all
-all: prepare lint test_rules
+all: prepare lint test_rules pint_lint
 
 .PHONY: prepare
-prepare:
+prepare: pint
 	./automation/promtool_tests.sh prepare
 
 .PHONY: lint
@@ -13,6 +15,13 @@ lint:
 test_rules:
 	./automation/promtool_tests.sh test_rules
 
-.PHONY: pint
 pint:
+	echo 'Installing pint...'
+	curl -OJL 'https://github.com/cloudflare/pint/releases/download/v${PINT_VERSION}/pint-${PINT_VERSION}-linux-amd64.tar.gz'
+	tar xvzf 'pint-${PINT_VERSION}-linux-amd64.tar.gz' pint-linux-amd64
+	mv pint-linux-amd64 pint
+	rm 'pint-${PINT_VERSION}-linux-amd64.tar.gz'	
+	
+.PHONY: pint_lint
+pint_lint:	
 	./automation/pint_lint.sh
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 promtool
 yq
 test/promql/extracted-rules.yaml
+pint
--- a/.tekton/pull-request.yaml
+++ b/.tekton/pull-request.yaml
@@ -94,7 +94,7 @@ spec:
             - name: lint-rules
               image: registry.access.redhat.com/ubi8/ubi:latest
               workingDir: $(workspaces.workspace.path)
-              script: yum install -y make && make pint
+              script: yum install -y make && make pint_lint
       - name: run-gitlint
         params:
           - name: target-branch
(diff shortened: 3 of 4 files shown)


Omeramsc (approved):
LGTM

Makefile
```diff
@@ -13,6 +15,9 @@ lint:
 test_rules:
 	./automation/promtool_tests.sh test_rules
 
-.PHONY: pint
 pint:
-	./automation/pint_lint.sh
```
yftacherzog, line 18: Are we still running the tests?

eisraeli: @yftacherzog There's still an issue here I'm trying to resolve..

yftacherzog: Yes. I think we need them separated into 2 targets. One will be called `pint` and will install it and another will be phony and will have a different name (I'm sure that there are ways to tell make to consider binaries with different names than the target name, but I'm not familiar enough with that).

I'd say the installation needs to be a target which will be a dependency for the `prepare` target (later on we can have the other parts of `prepare` as targets which will be dependencies for `prepare` as well), and then running the tests can happen as another target (with a name which is not `pint`).

yftacherzog: Something like this:
```
all: prepare lint test_rules run_pint

prepare: pint
     ./automation/promtool_tests.sh prepare (we'll change this on a later PR as well and turn it into another dependency)

pint:
	echo 'Installing pint...'
	curl -OJL 'https://github.com/cloudflare/pint/releases/download/v${PINT_VERSION}/pint-${PINT_VERSION}-linux-amd64.tar.gz'
	tar xvzf 'pint-${PINT_VERSION}-linux-amd64.tar.gz' pint-linux-amd64
	mv pint-linux-amd64 pint
	rm 'pint-${PINT_VERSION}-linux-amd64.tar.gz'

.PHONY:  run_pint
run_pint:
        ./automation/pint_lint.sh
```

Makefile
```diff
@@ -1,8 +1,11 @@
+PINT_VERSION='0.42.2'
+
 .PHONY: all
-all: prepare lint test_rules
+all: prepare lint test_rules pint_lint
 
 .PHONY: prepare
 prepare:
+	make pint
```
yftacherzog, line 8: I think it's better to have pint as a dependency to prepare rather than invoking `make` again.
```suggestion
prepare: pint
```

Makefile
```diff
@@ -13,6 +16,15 @@ lint:
 test_rules:
 	./automation/promtool_tests.sh test_rules
 
-.PHONY: pint
+.PHONY: test_rules
```
yftacherzog, line 19: We already have this line earlier. I don't think we need it again.

Makefile
yftacherzog: Temporary in what sense? You mean until we make it all part of an image?

eisraeli: @yftacherzog Yup

yftacherzog: :crossed_fingers:

Makefile
```diff
@@ -13,6 +15,13 @@ lint:
...
-.PHONY: pint
 pint:
+	echo 'Installing pint...'
+	curl -OJL 'https://github.com/cloudflare/pint/releases/download/v${PINT_VERSION}/pint-${PINT_VERSION}-linux-amd64.tar.gz'
+	tar xvzf 'pint-${PINT_VERSION}-linux-amd64.tar.gz' pint-linux-amd64
+	mv pint-linux-amd64 pint
+	rm 'pint-${PINT_VERSION}-linux-amd64.tar.gz'	
+	
+.PHONY: pint_lint
+pint_lint:
```
yftacherzog, line 26: You need to modify the call to make pull-request.yaml so it refer to this target. Currently, it just runs the `pint` target again.
L97 there should be:
```
script: yum install -y make && make pint_lint
```

amisstea (approved):

yftacherzog (approved):
I think we're good :)

PR #25, opened 2023-03-24
test(STONEO11Y-77): add yamllint to local tests

Kousalya1998:
- Added installation scripts of `yamllint` in a Python virtual env to Makefile
- Added yaml linting to Makefile
- Added Python virtual env (`.venv`) to .gitignore


eisraeli:
Hi,
Notice the checks on the PR.
There are still errors.

**prepare-for-prometheus-tests:**
`make: python3: Command not found
make: *** [Makefile:54: yamllint] Error 127`

**run-gitlint**
```
1: CC1 Body does not contain a 'Signed-off-by' line
3: B6 Body message is missing
```

.tekton/pull-request.yaml
```diff
@@ -95,6 +95,20 @@ spec:
               image: registry.access.redhat.com/ubi8/ubi:latest
               workingDir: $(workspaces.workspace.path)
               script: yum install -y make && make pint_lint
+      - name: lint-prometheus-rules-yaml-lint
+        workspaces:
+          - name: workspace
+            workspace: workspace
+        runAfter:
+          - prepare-for-prometheus-tests
```
eisraeli, line 103: It doesn't have to run after `prepare-for-prometheus-tests` 
Actually, I would exclude the Prometheus rules files as they are already linted by promtools and pint.

.gitignore
```diff
@@ -2,3 +2,4 @@ promtool
 yq
 test/promql/extracted-rules.yaml
 pint
+.venv
```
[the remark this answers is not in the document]
Kousalya1998: > Notice that you are missing a newline at the end of the file

Added a newline

yftacherzog (changes requested):
@Kousalya1998 kudos on creating this PR early on!

The idea behind this ticket is to consolidate the way yamllint is executed in CI and in local dev envs. This means a couple of things:

1. have the yamllint procedure run from within the makefile.
2. Have the CI trigger the same makefile target that will be triggered by the developers.

We also need to make sure CI and devs use the same yamllint version. The standard way to go about it is to use pipenv and then have all environments use it to get everything installed with the same versions.

Seems as the image we're currently using doesn't have pipenv installed. You can have it installed with `pip install pipenv`, then you will need to activate the virtual environment. It might be that you'd need to use some flags to make it ignore the existing virtualenv that this image supplies (we can try without it at first. not sure it's a problem to use the supplied one).

All in all, in your local setup, you'd need to generate the pipfile and lock by installing yamllint with pipenv (`pipenv install yamllint`). Those 2 files need to go inside git. Then the CI will have to install pipenv, activate the pipenv (`pipenv sync --dev`) and run yamllint.

Need to figure out exactly which parts of that only happen in CI and which parts need to happen on makefile targets that also run in dev envs.

.tekton/pull-request.yaml
Kousalya1998: > It doesn't have to run after `prepare-for-prometheus-tests` Actually, I would exclude the Prometheus rules files as they are already linted by promtools and pint and the extraction of the rules is not relevant for yamllint (promtools and pint can't run without the extraction though)

Excluded those lines. Thanks

.gitignore
```diff
@@ -2,3 +2,5 @@ promtool
...
```
yftacherzog, line 5: I'm not sure we need this one

.gitignore
```diff
@@ -2,3 +2,5 @@ promtool
...
+Pipfile.lock
```
yftacherzog, line 6: This must NOT be in gitignore. The idea is that we're locking the dependencies and then all envs sync according to the locked dependencies.

CONTRIBUTING.md
```diff
@@ -13,6 +13,7 @@ This section describes the main steps required before you can add your code and
 ### Prerequisites
 * Linux machine with x86-64 architecture with bash installed.
 * Basic tools: curl, tar, make.
+* Python3 with pipenv package.
```
yftacherzog, line 16: I think we will also need the step of syncing with the lock file. Need to see if this can be done as part of the makefile in a way that can be run multiple times and work correctly and not make execution take too long.

Makefile
```diff
@@ -46,3 +46,13 @@ yq:
 pint_lint:
 	echo "Linting Prometheus rules..."
 	./pint --no-color lint ${EXTRACTED_RULE_FILE}
+
+yamllint:
+	echo "Installing yamllint..."
+	pip install pipenv
+	pipenv sync --dev --skip-lock
```
yftacherzog, line 53: We should just sync with an existing lock.
In any case, I don't think it will help us to have it run here. We cannot run it in advance.
The CI is running each of the steps of the pull-request.yaml on a different image. They share the working directory but nothing else. It means, that we can fetch files on one step and use it on the other, but we cannot load a python virtual env to memory on one container and use it on another.

All in all, at least in CI, we must load the virtual env and use it on the same step. It can be on different makefile targets if it will help us sharing steps as much as possible between dev and ci, but it cannot happen on different steps.

.tekton/pull-request.yaml
```diff
@@ -50,7 +50,7 @@ spec:
             - name: workspace
           steps:
             - name: run-tests
-              image: registry.access.redhat.com/ubi8/ubi:latest
+              image: registry.access.redhat.com/ubi9/python-39:latest
```
yftacherzog, line 53: We should not change this image. It won't make a difference. Our changes should not affect this step at all - not on this yaml and not on the makefile as far as I can see it.

.tekton/pull-request.yaml
yftacherzog, line 1: The change that you'd need to this file is something like this (without any changes to other parts as far as I can see):
```
      - name: yaml-lint
        workspaces:
          - name: workspace
            workspace: workspace
        runAfter:
          - clone-repository
        taskSpec:
          workspaces:
            - name: workspace
          steps:
            - name: yaml-lint
              image: registry.access.redhat.com/ubi9/python-39:latest
              workingDir: $(workspaces.workspace.path)
              script: make install_pipenv sync_pipenv lint_yamls
```

Then on the makefile you will need to implement those targets:

```
 .PHONY: install_pipenv
install_pipenv:
	pip install pipenv

.PHONY: sync_pipenv
sync_pipenv:
	pipenv sync --dev

.PHONY: lint_yamls
lint_yamls:
	yamllint .
```

Then on the `all` target, we can debate on whether we just want to call `lint_yamls` or we need something else before that. i.e. do we make the part of going into the virtualenv something that the developers will do manually or would it be part of the running `make all` in some way.

I think we can try to include it at first to make things more streamlined. If it causes issues then we can take it out of the makefile and move it to the contributing guide :)

.gitignore
Kousalya1998: > I'm not sure we need this one

Removed from  gitignore

.gitignore
Kousalya1998: > This must NOT be in gitignore. The idea is that we're locking the dependencies and then all envs sync according to the locked dependencies.

Thank you for your explanation. It has been removed from gitgnore and the Pipfile.lock has been updated with all the required versions.

CONTRIBUTING.md
Kousalya1998: > Need to see if we will also need the step of syncing with the lock file. Need to see if this can be done as part of the makefile in a way that can be run multiple times and work correctly and not make execution take too long.

Since this prerequisite is not necessary at the moment, I removed it from the contributing guide and updated it.

Makefile
```diff
@@ -5,7 +5,7 @@ EXTRACTED_RULE_FILE = test/promql/extracted-rules.yaml
 RULE_FILE = prometheus/base/prometheus.rules.yaml
 
 .PHONY: all
-all: prepare lint test_rules pint_lint
+all: prepare install_pipenv sync_pipenv lint test_rules pint_lint lint_yamls
```
yftacherzog, line 8: Installing pipenv is a prerequisite you put on the contributing guide so you don't need to include it in make all. It will just run in ci.

Makefile
```diff
@@ -46,3 +46,25 @@ yq:
...
+
+yamllint:
+	echo "Installing yamllint..."
+	python3 -m pip install pipenv
+	python3 -m pipenv install --dev
+
+.PHONY: yaml_lint
+yaml_lint:
+	echo "Linting YAML files..."
+	python3 -m pipenv run yamllint . && echo "yaml_lint SUCCESS"
```
yftacherzog, line 58: You now have two targets with the same content. You can remove one.

yftacherzog (changes requested):
Nice progress!

.tekton/pull-request.yaml
Kousalya1998: Having understood what went wrong here, I have updated and modified the code accordingly. All test cases have now been passed.

.tekton/pull-request.yaml
Kousalya1998: > Nice progress!

Thank you for your inputs.

Makefile
Kousalya1998: I think it's best if I remove it from the contributing guide and not modify the makefile, because it will helpful for the developers to just run one command ‘make all’. In that way, it will be one less step for the developers. Whats your suggestion?

yftacherzog: I'd prefer the makefile not to hide installing of packages that has effect on the user's machine outside of the local repo and virtual env. 

@eisraeli @amisstea @avi-biton @hmariset any thoughts?

CONTRIBUTING.md
```diff
@@ -13,6 +13,7 @@ This section describes the main steps required before you can add your code and
 ### Prerequisites
 * Linux machine with x86-64 architecture with bash installed.
 * Basic tools: curl, tar, make.
+* Python3 and pipenv is required.
```
yftacherzog, line 16:
```suggestion
* Python3 and pipenv
```

Makefile
yftacherzog: You marked this as resolved but you still have both `yaml_lint` and `lint_yamls` doing the same. `yaml_lint` is not being referenced, so I think it should go.

Makefile
```diff
@@ -46,3 +46,25 @@ yq:
 pint_lint:
 	echo "Linting Prometheus rules..."
 	./pint --no-color lint ${EXTRACTED_RULE_FILE}
+
+yamllint:
```
yftacherzog, line 50: Can this target be deleted?

Makefile
yftacherzog: @Omeramsc

Makefile
```diff
@@ -46,3 +46,25 @@ yq:
...
+install_pipenv:
+	python3 -m pip install pipenv
+
+.PHONY: sync_pipenv
+sync_pipenv:
+	python3 -m pipenv sync --dev
+
+.PHONY: lint_yamls
+lint_yamls:
+	python3 -m pipenv run yamllint . && echo "lint_yamls: SUCCESS"
```
yftacherzog, line 70: Did you verify that this can also fail? i.e. temporarily add a yaml that should fail yamllint and verify that ci fails in the way it should.

Makefile
eisraeli: I agree with @yftacherzog 
Please don't change anything on the user's machine outside of the  local repo and virtualenv.  It's too intrusive.

Makefile
Kousalya1998: Deleted

CONTRIBUTING.md
Kousalya1998: Updated the CONTRIBUTING doc.

Makefile
Kousalya1998: Thank you for highlighting it, it has been removed.

Makefile
Kousalya1998: I understand, I have removed the `installing pipenv` from the Makefile.

> I agree with @yftacherzog Please don't change anything on the user's machine outside of the local repo and virtualenv. It's too intrusive.

Got it, I have removed the `installing pipenv` from the Makefile.

Makefile
Kousalya1998: Verified now, it fails in the CI too.

yftacherzog (approved):

srivickynesh (approved):
LGTM

hmariset (approved):
lgtm

PR #26, opened 2023-03-29
fix(STONEO11Y-86): change namespace on which prom rules are deployed

yftacherzog:
Prometheus rules were deployed on namespace o11y, but this way they are not being loaded by Prometheus. Here we change the namespace to be openshift-user-workload-monitoring so that they are picked up and evaluated


yftacherzog:
After researching into this a bit, it seems as the prometheus instance to capture rules under namespace `openshift-user-workload-monitoring` is actually the non-user-workload prometheus. Still need to figure out what needs to change so we capture rules with the correct instance.

PR #29, opened 2023-04-03
chore(STONEO11Y-117): Update CONTRIBUTING guide

Kousalya1998:
* Updated  CONTRIBUTING.md with pipenv installation command.


CONTRIBUTING.md
```diff
@@ -13,7 +13,7 @@ This section describes the main steps required before you can add your code and
 ### Prerequisites
 * Linux machine with x86-64 architecture with bash installed.
 * Basic tools: curl, tar, make.
-* Python3 and pipenv.
+* Python3 and pipenv. (eg: `python3 -m pip install pipenv --user`)
```
yftacherzog, line 16: I'd avoid providing it as an example here, as we already know that alternative approaches for installing pipenv might not work. I suggest instead mentioning what needs to be run. Something like:
```suggestion
* Python3 and pipenv. (run: `python3 -m pip install pipenv --user`)
```

Kousalya1998: Makes sense. I updated it now.

avi-biton:
LGTM

avi-biton (approved):
Please edit your commit message to pass the tests

srivickynesh (approved):
LGTM

PR #30, opened 2023-04-04
test(STONEO11Y-116): kustomize build passes

yftacherzog:
Add a test that verifies the configurations we have can be built with kustomize. This will prevent us from committing configs that will later fail deployment.

We can later on add a test that verifies the actual content of the yamls being generated. I wanted to add something simple to make sure the configs are buildable.

Making yaml comparisons is less trivial as it will require some sort of comparison that will not fail over ordering issues etc. I'm sure it can be done, but it will take some more effort in validation and I don't want to introduce flakiness hurrying to get this in.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ promtool
 yq
 test/promql/extracted-rules.yaml
 pint
+kustomize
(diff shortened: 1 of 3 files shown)


hmariset (approved):
LGTM

Kousalya1998 (approved):

Kousalya1998 (approved):
lgtm

PR #33, opened 2023-04-17
chore(STONEO11Y-101): Alerting rule for Quota Exceeded

ShakedAviv50313:
Create an alerting rule for Quota Exceeded error and add related unit test cases.

### DIFF
--- a/prometheus/base/alerting/prometheus.quota_alerts.yaml
+++ b/prometheus/base/alerting/prometheus.quota_alerts.yaml
@@ -1,0 +1,22 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    openshift.io/prometheus-rule-evaluation-scope: leaf-prometheus
+  name: o11y-quota-alerting-rules
+  namespace: o11y
+spec:
+  groups:
+  - name: quota_alerts
+    rules:
+    - alert: QuotaExceeded
+      expr: |
+        kube_resourcequota{type="used", namespace!~"(.*-tenant|openshift-.*|kube-.*|default)"} /
+        ignoring (type)
+        (kube_resourcequota{type="hard", namespace!~"(.*-tenant|openshift-.*|kube-.*|default)"} > 0) > 1
+      for: 15m
+      labels:
+        severity: warning
+      annotations:
+        summary: "Resource {{ $labels.resource }} in namespace {{ $labels.namespace }} exceeded quota {{ $labels.resourcequota }}"
+        description: "Resource {{ $labels.resource }} in namespace {{ $labels.namespace }} exceeded quota {{ $labels.resourcequota }}"
--- a/test/promql/tests/test_quota_exceeded.yaml
+++ b/test/promql/tests/test_quota_exceeded.yaml
@@ -1,0 +1,38 @@
+evaluation_interval: 1m
+
+rule_files:
+  - ../extracted-rules.yaml
+
+tests:
+  - interval: 1m
+    input_series:
+
+      # namespace does not reach its quota limits
+      - series: 'kube_resourcequota{type="hard", namespace="not-reaching-limits", resourcequota="test-quota", resource="example-resource"}'
+        values: '2x15'
+      - series: 'kube_resourcequota{type="used", namespace="not-reaching-limits", resourcequota="test-quota", resource="example-resource"}'
+        values: '0x15'
+
+      # namespace reaches its quota limits
+      - series: 'kube_resourcequota{type="hard", namespace="reaching-limits", resourcequota="test-quota", resource="example-resource"}'
+        values: '2x15'
+      - series: 'kube_resourcequota{type="used", namespace="reaching-limits", resourcequota="test-quota", resource="example-resource"}'
+        values: '2x15'
+
+      # namespace exceeds its quota limits
+      - series: 'kube_resourcequota{type="hard", namespace="exceeding-limits", resourcequota="test-quota", resource="example-resource"}'
+        values: '2x15'
+      - series: 'kube_resourcequota{type="used", namespace="exceeding-limits", resourcequota="test-quota", resource="example-resource"}'
+        values: '5x15'
+    alert_rule_test:
+      - eval_time: 15m
+        alertname: QuotaExceeded
+        exp_alerts:
+          - exp_labels:
+              severity: warning
+              namespace: exceeding-limits
+              resource: example-resource
+              resourcequota: test-quota
+            exp_annotations:
+              summary: "Resource example-resource in namespace exceeding-limits exceeded quota test-quota"
+              description: "Resource example-resource in namespace exceeding-limits exceeded quota test-quota"
--- a/prometheus/base/alerting/kustomization.yaml
+++ b/prometheus/base/alerting/kustomization.yaml
@@ -2,3 +2,4 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
 - prometheus.pod_alerts.yaml
+- prometheus.quota_alerts.yaml


prometheus/base/alerting/prometheus.pod_alerts.yaml
```diff
@@ -25,3 +25,14 @@ spec:
       annotations:
         summary: "Pod {{ $labels.pod }} is crash looping"
         description: "Pod {{ $labels.namespace }}/{{ $labels.pod }} ({{ $labels.container }}) is in waiting state (reason: 'CrashLoopBackOff') for more than 15 minutes."
+    - alert: QuotaExceeded
+      expr: |
+        kube_resourcequota{type="used", namespace!~"(.*-tenant|openshift-.*|kube-.*|default)"} /
+        ignoring (instance, job, type)
+        (kube_resourcequota{type="hard", namespace!~"(.*-tenant|openshift-.*|kube-.*|default)"} > 0) > 1
```
yftacherzog, line 32: @amisstea is this ever going to be higher than one?

test/promql/tests/test_quota_exceeded.yaml
yftacherzog, line 1: To make the tests more realistic, I suggest also adding a resource type. You could then also include that resource in the alert message. Something like this:

```
      - series: 'kube_resourcequota{type="hard", namespace="not-reaching-limits", resourcequota="some-quota", resource="some-resource"}'
```

yftacherzog: I don't have an example in front of me, but see which labels you're getting from `kube_resourcequota` and see if including any of them on the alert rule message will be helpful for the receiver of alert.

prometheus/base/alerting/prometheus.pod_alerts.yaml
```diff
@@ -25,3 +25,14 @@ spec:
       annotations:
         summary: "Pod {{ $labels.pod }} is crash looping"
         description: "Pod {{ $labels.namespace }}/{{ $labels.pod }} ({{ $labels.container }}) is in waiting state (reason: 'CrashLoopBackOff') for more than 15 minutes."
+    - alert: QuotaExceeded
+      expr: |
+        kube_resourcequota{type="used", namespace!~"(.*-tenant|openshift-.*|kube-.*|default)"} /
+        ignoring (instance, job, type)
```
yftacherzog, line 31: @amisstea do you recall what was the idea behind ignoring instance and job?

Omeramsc:
Please include a description for the PR itself

test/promql/tests/test_quota_exceeded.yaml
avi-biton: In continue to that, we have the `resource` label that we can use to better describe what resource quota is exceeded

prometheus/base/alerting/prometheus.pod_alerts.yaml
```diff
@@ -25,3 +25,14 @@ spec:
...
+        (kube_resourcequota{type="hard", namespace!~"(.*-tenant|openshift-.*|kube-.*|default)"} > 0) > 1
+      for: 15m
+      labels:
+        severity: warning
+      annotations:
+        summary: "Namespace {{ $labels.namespace }} exceeded its quota"
```
avi-biton, line 37: We have the resource label that we can use to better describe what resource quota is exceeded

Kousalya1998:
Please add a description for this PR.

prometheus/base/alerting/prometheus.pod_alerts.yaml
amisstea: I'm not quite sure. I don't see `instance` as an available label on this metric and `job` is always the same value from all samples I've seen. Perhaps the implementers of the platform rule found some edge cases where these must be ignored. I suggest we ignore only `type` for now and revisit if it becomes a problem. Let's not carry forward potential tech debt we can't explain.

prometheus/base/alerting/prometheus.pod_alerts.yaml
amisstea: Yes. I experimented by creating a `Pod` and then a `ResourceQuota` with limits and requests lower than what was defined in the `Pod`.

image

test/promql/tests/test_quota_exceeded.yaml
amisstea: The `container`, `resource` and `resourcequota` labels look like they would be potentially useful here.

ShakedAviv50313: I fixed it, now the tests contain also a ```resource``` and a ```resourcequota```.
The summary and the description now also contain the ```namespace```,  the ```resource``` and the ```resourcequota```

prometheus/base/alerting/prometheus.pod_alerts.yaml
ShakedAviv50313: @avi-biton How about now?

prometheus/base/alerting/prometheus.pod_alerts.yaml
yftacherzog: Cool

yftacherzog: On second thought, suppose that we consider the more standard flow in which a quota is created before the pod, will it ever alert? So maybe it should be higher or equal instead?

and also, is it possible that the quota will be 98% but resources cannot be created, but the will not fill the quota?
Suppose you have 10GB mem quota and its 9GB taken, and you try to create 2GB pods, will the quota reach 100%? Should we set the value to something smaller than 100%?

prometheus/base/alerting/prometheus.pod_alerts.yaml
yftacherzog: @ShakedAviv50313 can you adjust it accordingly?
```suggestion
        ignoring (type)
```

prometheus/base/alerting/prometheus.pod_alerts.yaml
yftacherzog, line 1: I think this alert should be moved away from the pod alerts file, as it's not related to pod metrics.

prometheus/base/alerting/prometheus.pod_alerts.yaml
ShakedAviv50313: Done

prometheus/base/alerting/prometheus.pod_alerts.yaml
ShakedAviv50313: Done

prometheus/base/alerting/prometheus.quota_alerts.yaml
```diff
@@ -0,0 +1,22 @@
...
+      expr: |
+        kube_resourcequota{type="used", namespace!~"(.*-tenant|openshift-.*|kube-.*|default)"} /
+        ignoring (type)
+        (kube_resourcequota{type="hard", namespace!~"(.*-tenant|openshift-.*|kube-.*|default)"} > 0) > 1
+      for: 15m
+      labels:
+        severity: warning
+      annotations:
+        summary: "Resource {{ $labels.resource }} in namespace {{ $labels.namespace }} exceeded quota {{ $labels.resourcequota }}"
+        description: "Resource {{ $labels.resource }} in namespace {{ $labels.namespace }} exceeded quota {{ $labels.resourcequota }}"
```
yftacherzog, line 22: add newline at end of file

prometheus/base/alerting/prometheus.pod_alerts.yaml
yftacherzog: Need to also update kustomization.yaml so we build the new resource

prometheus/base/alerting/prometheus.quota_alerts.yaml
yftacherzog: I think now you have a line with some whitespaces at the end of the file. You need it to be an empty line.

prometheus/base/alerting/prometheus.pod_alerts.yaml
ShakedAviv50313: also done

prometheus/base/alerting/prometheus.quota_alerts.yaml
ShakedAviv50313: fixed it, it is now an empty line

prometheus/base/alerting/prometheus.pod_alerts.yaml
amisstea: Also, if we do want to fire alerts for high quota usage, I think it should be defined as a separate alert.

amisstea (approved):

PR #34, opened 2023-04-17
chore(STONEO11Y-96): Alerting rule for high controller reconciliation errors

hmariset:
Create an alerting rule for controller reconciliation errors and add related unit test cases.

### DIFF
--- a/prometheus/base/alerting/prometheus.controller_alerts.yaml
+++ b/prometheus/base/alerting/prometheus.controller_alerts.yaml
@@ -1,0 +1,23 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    openshift.io/prometheus-rule-evaluation-scope: leaf-prometheus
+  name: o11y-controller-alerting-rules
+  namespace: o11y
+spec:
+  groups:
+  - name: controller_alerts
+    rules:
+    - alert: ControllerReconciliationErrors
+      expr: |
+        sum(increase(controller_runtime_reconcile_errors_total{namespace!~"(.*-tenant|openshift-.*|kube-.*|default)"}[2m]))
+        by(container, controller, namespace, pod, service) /
+        sum(increase(controller_runtime_reconcile_total{namespace!~"(.*-tenant|openshift-.*|kube-.*|default)"}[2m]))
+        by(container, controller, namespace, pod, service) > 0.01
+      for: 2m
+      labels:
+        severity: warning
+      annotations:
+        summary: "Controller {{ $labels.controller }} has reconciliation errors greater than 1%"
+        description: "Controller {{ $labels.controller }} with Pod {{ $labels.pod }} for namespace {{ $labels.namespace }} has total reconciliation errors greater than 1% compared to total reconciliations in last 2 minutes."
--- a/test/promql/tests/test_controller_reconciliation_errors.yaml
+++ b/test/promql/tests/test_controller_reconciliation_errors.yaml
@@ -1,0 +1,127 @@
+evaluation_interval: 1m
+
+rule_files:
+  - ../extracted-rules.yaml
+
+tests:
+
+  # Alerted test cases
+  - interval: 1m
+    input_series:
+      # Controller with continuous errors greater than 1% so this is alerted
+      - series: 'controller_runtime_reconcile_errors_total{controller="c1-controller", namespace="prod-ns1", pod="pod-1", container="c1", service="s1"}'
+        values: '1 2 3'
+      - series: 'controller_runtime_reconcile_total{controller="c1-controller", namespace="prod-ns1", pod="pod-1", container="c1", service="s1"}'
+        values: '1 2 3'
+
+    alert_rule_test:
+      - eval_time: 3m
+        alertname: ControllerReconciliationErrors
+        exp_alerts:
+          - exp_labels:
+              severity: warning
+              namespace: prod-ns1
+              pod: pod-1
+              controller: c1-controller
+              container: c1
+              service: s1
+            exp_annotations:
+              summary: "Controller c1-controller has reconciliation errors greater than 1%"
+              description: "Controller c1-controller with Pod pod-1 for namespace prod-ns1 has total reconciliation errors greater than 1% compared to total reconciliations in last 2 minutes."
+
+  - interval: 1m
+    input_series:
+      # Controller without error at start and then error starts. This is alerted because there is always an increase in error value
+      - series: 'controller_runtime_reconcile_errors_total{controller="c2-controller", namespace="prod-ns2", pod="pod-2", container="c2", service="s2"}'
+        values: '0 1 2'
+      - series: 'controller_runtime_reconcile_total{controller="c2-controller", namespace="prod-ns2", pod="pod-2", container="c2", service="s2"}'
+        values: '1 2 3'
+
+    alert_rule_test:
+      - eval_time: 3m
+        alertname: ControllerReconciliationErrors
+        exp_alerts:
+          - exp_labels:
+              severity: warning
+              namespace: prod-ns2
+              pod: pod-2
+              controller: c2-controller
+              container: c2
+              service: s2
+            exp_annotations:
+              summary: "Controller c2-controller has reconciliation errors greater than 1%"
+              description: "Controller c2-controller with Pod pod-2 for namespace prod-ns2 has total reconciliation errors greater than 1% compared to total reconciliations in last 2 minutes."
+
+  - interval: 1m
+    input_series:
+      # Controller with large spikes in error values and pause in reconcile total,
+      # since the error rate is always greater than 1% in 2min window this is alerted
+      - series: 'controller_runtime_reconcile_errors_total{controller="c10-controller", namespace="prod-ns10", pod="pod-10", container="c10", service="s10"}'
+        values: '1 150 390'
+      - series: 'controller_runtime_reconcile_total{controller="c10-controller", namespace="prod-ns10", pod="pod-10", container="c10", service="s10"}'
+        values: '1 500 740'
+
+    alert_rule_test:
+      - eval_time: 3m
+        alertname: ControllerReconciliationErrors
+        exp_alerts:
+          - exp_labels:
+              severity: warning
+              namespace: prod-ns10
+              pod: pod-10
+              controller: c10-controller
+              container: c10
+              service: s10
+            exp_annotations:
+              summary: "Controller c10-controller has reconciliation errors greater than 1%"
+              description: "Controller c10-controller with Pod pod-10 for namespace prod-ns10 has total reconciliation errors greater than 1% compared to total reconciliations in last 2 minutes."
+
+  # Not Alerted test cases
+  - interval: 1m
+    input_series:
+      # Controller with no reconciliation events throughout the test time frame so this is not alerted
+      - series: 'controller_runtime_reconcile_errors_total{controller="c5-controller", namespace="prod-ns5", pod="pod-5", container="c1", service="s1"}'
+        values: '1 1 1'
+      - series: 'controller_runtime_reconcile_total{controller="c5-controller", namespace="prod-ns5", pod="pod-5", container="c1", service="s1"}'
+        values: '10 10 10'
+
+      # Controller with errors during parts of the test period
+      - series: 'controller_runtime_reconcile_errors_total{controller="c3-controller", namespace="prod-ns3", pod="pod-3", container="c1", service="s1"}'
+        values: '0 2 2'
+      - series: 'controller_runtime_reconcile_total{controller="c3-controller", namespace="prod-ns3", pod="pod-3", container="c1", service="s1"}'
+        values: '1 3 3'
+
+      # Controller with error throughout but their total sum is less than 1% of the total number of
+      # reconciliation events at the test period so this is not alerted
+      - series: 'controller_runtime_reconcile_errors_total{controller="c4-controller", namespace="prod-ns4", pod="pod-4", container="c1", service="s1"}'
+        values: '1 2 3'
+      - series: 'controller_runtime_reconcile_total{controller="c4-controller", namespace="prod-ns4", pod="pod-4", container="c1", service="s1"}'
+        values: '1 1000 1001'
+
+      # Controller with error throughout but namespace starts with openshift so this is not alerted
+      - series: 'controller_runtime_reconcile_errors_total{controller="c7-controller", namespace="openshift-ns", pod="pod-1", container="c1", service="s1"}'
+        values: '1 2 3'
+      - series: 'controller_runtime_reconcile_total{controller="c7-controller", namespace="openshift-ns", pod="pod-1", container="c1", service="s1"}'
+        values: '1 2 3'
+
+      # Controller with error throughout but namespace starts with kube so this is not alerted
+      - series: 'controller_runtime_reconcile_errors_total{controller="c8-controller", namespace="kube-ns", pod="pod-1", container="c1", service="s1"}'
+        values: '1 2 3'
+      - series: 'controller_runtime_reconcile_total{controller="c8-controller", namespace="kube-ns", pod="pod-1", container="c1", service="s1"}'
+        values: '1 2 3'
+
+      # Controller with error throughout but namespace ends with tenant so this is not alerted
+      - series: 'controller_runtime_reconcile_errors_total{controller="c9-controller", namespace="ns-tenant", pod="pod-1", container="c1", service="s1"}'
+        values: '1 2 3'
+      - series: 'controller_runtime_reconcile_total{controller="c9-controller", namespace="ns-tenant", pod="pod-1", container="c1", service="s1"}'
+        values: '1 2 3'
+
+      # Controller with error throughout but has default as namespace so this is not alerted
+      - series: 'controller_runtime_reconcile_errors_total{controller="c0-controller", namespace="default", pod="pod-1", container="c1", service="s1"}'
+        values: '1 2 3'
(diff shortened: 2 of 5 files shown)


test/promql/tests/test_controller_reconciliation_errors.yaml
```diff
@@ -0,0 +1,75 @@
+evaluation_interval: 1m
+
+rule_files:
+  - ../extracted-rules.yaml
+
+tests:
+  - interval: 1m
```
yftacherzog, line 7: i suggest splitting this to multiple test cases. Have one test case for each alerting pod, and then we can have another test case for all the non-alerting pods.

yftacherzog:
> … errors
> 
> Create an alerting rule for controller reconciliation errors and add related unit test cases.

Can you fix the PR description?

prometheus/base/alerting/prometheus.controller_alerts.yaml
```diff
@@ -0,0 +1,23 @@
...
+  labels:
+    openshift.io/prometheus-rule-evaluation-scope: leaf-prometheus
+  name: o11y-controller-alerting-rules
+  namespace: o11y
+spec:
+  groups:
+  - name: controller_alerts
+    rules:
+    - alert: ControllerReconciliationErrors
+      expr: |
```
yftacherzog, line 13: The requirement as I understand it from @amisstea's document was involving a lookback of 2m window. I don't see this here.

https://docs.google.com/document/d/1akkvRQRiQud-ba644bSrqhpq1asJ4H-YO_DA58BjE-s/edit#heading=h.20eg8stzvxhy

prometheus/base/alerting/prometheus.controller_alerts.yaml
```diff
@@ -0,0 +1,23 @@
...
+        sum(controller_runtime_reconcile_errors_total{namespace!~"(.*-tenant|openshift-.*|kube-.*|default)"})
+        by(container, controller, namespace, pod, service) /
+        sum(controller_runtime_reconcile_total{namespace!~"(.*-tenant|openshift-.*|kube-.*|default)"})
+        by(container, controller, namespace, pod, service) * 100 > 1
+      for: 15m
+      labels:
+        severity: warning
+      annotations:
+        summary: "Controller {{ $labels.controller }} has errors greater than 1%"
```
Omeramsc, line 22: What does "errors greater than 1%" mean? 1% of what?

prometheus/base/alerting/prometheus.controller_alerts.yaml
hmariset: I need to check with @amisstea on this since it's not alerting if there is no change in values and error is still there.

test/promql/tests/test_controller_reconciliation_errors.yaml
```diff
@@ -0,0 +1,108 @@
...
+              summary: "Controller c1-controller has errors greater than 1%"
+              description: "Controller c1-controller with Pod pod-1 for namespace prod-ns1 has errors greater than 1% in last 15 minutes."
+
+  - interval: 1m
+    input_series:
+      # Controller with error throughout with value greater than 1% so this is alerted
+      - series: 'controller_runtime_reconcile_errors_total{controller="c5-controller", namespace="prod-ns5", pod="pod-5", container="c1", service="s1"}'
+        values: '1+1x14'
+      - series: 'controller_runtime_reconcile_total{controller="c5-controller", namespace="prod-ns5", pod="pod-5", container="c1", service="s1"}'
+        values: '10+1x14'
```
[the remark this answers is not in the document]
yftacherzog: I meant to say, what's the meaning of the increasing value.

prometheus/base/alerting/prometheus.controller_alerts.yaml
amisstea: I think the lookback window might be important or else this could end up detecting old controller errors and these alerts should only consider recent activity.

hmariset: @amisstea Thanks for the clarification. I have updated now.

prometheus/base/alerting/prometheus.controller_alerts.yaml
```diff
@@ -0,0 +1,23 @@
...
+spec:
+  groups:
+  - name: controller_alerts
+    rules:
+    - alert: ControllerReconciliationErrors
+      expr: |
+        sum(increase(controller_runtime_reconcile_errors_total{namespace!~"(.*-tenant|openshift-.*|kube-.*|default)"}[2m]))
+        by(container, controller, namespace, pod, service) /
+        sum(increase(controller_runtime_reconcile_total{namespace!~"(.*-tenant|openshift-.*|kube-.*|default)"}[2m]))
+        by(container, controller, namespace, pod, service) * 100 > 1
```
yftacherzog, line 17: I think it's better to eliminate the multiplication here
```suggestion
        by(container, controller, namespace, pod, service) > 0.01
```

test/promql/tests/test_controller_reconciliation_errors.yaml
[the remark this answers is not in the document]
yftacherzog: Maybe can change it to be constant, so it will not cause distraction.

test/promql/tests/test_controller_reconciliation_errors.yaml
```diff
@@ -0,0 +1,91 @@
...
+      - series: 'controller_runtime_reconcile_errors_total{controller="c3-controller", namespace="prod-ns3", pod="pod-3", container="c1", service="s1"}'
+        values: '1+1x6 0x5 1+1x6'
+      - series: 'controller_runtime_reconcile_total{controller="c3-controller", namespace="prod-ns3", pod="pod-3", container="c1", service="s1"}'
+        values: '1+1x19'
+
+      # Controller with error throughout but there is a window where the value is less than 1% so this is not alerted
+      - series: 'controller_runtime_reconcile_errors_total{controller="c4-controller", namespace="prod-ns4", pod="pod-4", container="c1", service="s1"}'
+        values: '1+1x19'
+      - series: 'controller_runtime_reconcile_total{controller="c4-controller", namespace="prod-ns4", pod="pod-4", container="c1", service="s1"}'
+        values: '1+1x7 1000x2 1+1x8'
```
yftacherzog, line 57: If errors are constantly increasing then the total number should increase as well - at least as much as the errors. Should be something like this:
```suggestion
        values: '1+1x7 1000+1x11'
```

test/promql/tests/test_controller_reconciliation_errors.yaml
```diff
@@ -0,0 +1,91 @@
...
+              description: "Controller c1-controller with Pod pod-1 for namespace prod-ns1 has total reconciliation errors greater than 1% compared to total reconciliations in last 15 minutes."
+
+  # Not Alerted test cases
+  - interval: 1m
+    input_series:
+      # Controller with error throughout with value greater than 1% but it's constant series so it's not alerted
+      - series: 'controller_runtime_reconcile_errors_total{controller="c5-controller", namespace="prod-ns5", pod="pod-5", container="c1", service="s1"}'
+        values: '1x19'
+      - series: 'controller_runtime_reconcile_total{controller="c5-controller", namespace="prod-ns5", pod="pod-5", container="c1", service="s1"}'
+        values: '10x19'
```
amisstea, line 39: This simulates an idle controller to me. The error total will always be 1 and the reconcile total 10 so there's no activity over those 19m.

The test is useful but the description isn't clear.

test/promql/tests/test_controller_reconciliation_errors.yaml
```diff
@@ -0,0 +1,91 @@
...
+
+rule_files:
+  - ../extracted-rules.yaml
+
+tests:
+
+  # Alerted test cases
+  - interval: 1m
+    input_series:
+      # Controller with continuous errors greater than 1% so this is alerted
```
amisstea, line 11: Can you include some more variations of time series values such as large spikes of errors mixed with pauses in reconcile activity?

prometheus/base/alerting/prometheus.controller_alerts.yaml
```diff
@@ -0,0 +1,23 @@
...
+  groups:
+  - name: controller_alerts
+    rules:
+    - alert: ControllerReconciliationErrors
+      expr: |
+        sum(increase(controller_runtime_reconcile_errors_total{namespace!~"(.*-tenant|openshift-.*|kube-.*|default)"}[2m]))
+        by(container, controller, namespace, pod, service) /
+        sum(increase(controller_runtime_reconcile_total{namespace!~"(.*-tenant|openshift-.*|kube-.*|default)"}[2m]))
+        by(container, controller, namespace, pod, service) > 0.01
+      for: 15m
```
amisstea, line 18: I suspect by using 15m, it will fail to pickup bursts of errors (see my other comment in the tests). It would be nice to detect those of course. Setting this to `2m` might help but I haven't considered the possible downsides. Wdyt?

hmariset: @amisstea I have updated to 2m and accordingly test cases as well. Please review again.

test/promql/tests/test_controller_reconciliation_errors.yaml
```diff
@@ -0,0 +1,133 @@
...
+      - series: 'controller_runtime_reconcile_errors_total{controller="c1-controller", namespace="prod-ns1", pod="pod-1", container="c1", service="s1"}'
+        values: '1+1x2'
```
amisstea, line 13: Now that we only need a few samples, I think the shorthand notation used throughout these unit test series makes them less understandable than explicitly putting the values. I have to think far less if written like this:
```yml
values: '1 2 3'
```

test/promql/tests/test_controller_reconciliation_errors.yaml
```diff
@@ -0,0 +1,133 @@
...
+              container: c2
+              service: s2
+            exp_annotations:
+              summary: "Controller c2-controller has reconciliation errors greater than 1%"
+              description: "Controller c2-controller with Pod pod-2 for namespace prod-ns2 has total reconciliation errors greater than 1% compared to total reconciliations in last 2 minutes."
+
+  - interval: 1m
+    input_series:
+      # Controller with large spikes in error values and pause in reconcile total,
+      # since the error rate is always greater than 1% in 2min window this is alerted
```
yftacherzog, line 58: Can this actually happen? I was under the impression that `errors_total` is part of `total`. If that is true, then the increase in `total` should be at least the increase in `errors_total` (i.e. all added occurrences were errors). No?

yftacherzog: I'd imagine something like this:
```
      - series: 'controller_runtime_reconcile_errors_total{controller="c10-controller", namespace="prod-ns10", pod="pod-10", container="c10", service="s10"}'
        values: '1 150 390'
      - series: 'controller_runtime_reconcile_total{controller="c10-controller", namespace="prod-ns10", pod="pod-10", container="c10", service="s10"}'
        values: '1 500 740'
```

test/promql/tests/test_controller_reconciliation_errors.yaml
```diff
@@ -0,0 +1,133 @@
...
+              service: s10
+            exp_annotations:
+              summary: "Controller c10-controller has reconciliation errors greater than 1%"
+              description: "Controller c10-controller with Pod pod-10 for namespace prod-ns10 has total reconciliation errors greater than 1% compared to total reconciliations in last 2 minutes."
+
+  # Not Alerted test cases
+  - interval: 1m
+    input_series:
+      # Controller with error throughout but reconcile error and total is always constant with values 1 and 10
+      # even though the error rate is greater than 1% it's not alerted since the error is not increasing
```
yftacherzog, line 83: If I understand correctly, those two metrics are counters. Meaning, that if their values are constant, it means no activity. e.g. if error series is `1 1 1`, it means that there was an error some time before the first `1`, and then no errors.

I'm ok with the test case, but I think the comment should be different:
```suggestion
      # Controller with no reconciliation events throughout the test time frame.
```

test/promql/tests/test_controller_reconciliation_errors.yaml
```diff
@@ -0,0 +1,133 @@
...
+      - series: 'controller_runtime_reconcile_errors_total{controller="c5-controller", namespace="prod-ns5", pod="pod-5", container="c1", service="s1"}'
+        values: '1 1 1'
+      - series: 'controller_runtime_reconcile_total{controller="c5-controller", namespace="prod-ns5", pod="pod-5", container="c1", service="s1"}'
+        values: '10 10 10'
+
+      # Controller without error at start then the error starts but it's constant so this is not alerted
```
yftacherzog, line 89:
```suggestion
      # Controller with errors during parts of the test period.
```

test/promql/tests/test_controller_reconciliation_errors.yaml
```diff
@@ -0,0 +1,133 @@
...
+      - series: 'controller_runtime_reconcile_errors_total{controller="c3-controller", namespace="prod-ns3", pod="pod-3", container="c1", service="s1"}'
+        values: '0 2 2'
+      - series: 'controller_runtime_reconcile_total{controller="c3-controller", namespace="prod-ns3", pod="pod-3", container="c1", service="s1"}'
+        values: '1 2 3'
+
+      # Controller with error throughout but there is a window where the value is less than 1% so this is not alerted
```
yftacherzog, line 95:
```suggestion
      # Controller with error throughout but their total sum is less than 1% of the total number of
      # reconciliation events at the test period so this is not alerted
```

test/promql/tests/test_controller_reconciliation_errors.yaml
```diff
@@ -0,0 +1,133 @@
...
+      - series: 'controller_runtime_reconcile_errors_total{controller="c4-controller", namespace="prod-ns4", pod="pod-4", container="c1", service="s1"}'
+        values: '1 2 3'
+      - series: 'controller_runtime_reconcile_total{controller="c4-controller", namespace="prod-ns4", pod="pod-4", container="c1", service="s1"}'
+        values: '1 1000 1001'
+
+      # Controller with error throughout but value is less than 1% so this is not alerted
```
yftacherzog, line 101: I'm not sure this is any different than the previous one.

test/promql/tests/test_controller_reconciliation_errors.yaml
hmariset: Yes, you are correct I thought connectivity issue with Kubernetes API wouldn't increase controller_runtime_reconcile_total but increases controller_runtime_reconcile_errors_total. But the other metric would also not increase due to that.

test/promql/tests/test_controller_reconciliation_errors.yaml
```diff
@@ -0,0 +1,127 @@
...
+      - series: 'controller_runtime_reconcile_errors_total{controller="c5-controller", namespace="prod-ns5", pod="pod-5", container="c1", service="s1"}'
+        values: '1 1 1'
+      - series: 'controller_runtime_reconcile_total{controller="c5-controller", namespace="prod-ns5", pod="pod-5", container="c1", service="s1"}'
+        values: '10 10 10'
+
+      # Controller with errors during parts of the test period
+      - series: 'controller_runtime_reconcile_errors_total{controller="c3-controller", namespace="prod-ns3", pod="pod-3", container="c1", service="s1"}'
+        values: '0 2 2'
+      - series: 'controller_runtime_reconcile_total{controller="c3-controller", namespace="prod-ns3", pod="pod-3", container="c1", service="s1"}'
+        values: '1 3 3'
```
amisstea, line 92: Why does this not raise the alert if during the first two minutes the increase in errors vs total is 2/2 or 100%?

prometheus/base/alerting/prometheus.controller_alerts.yaml
```diff
@@ -0,0 +1,23 @@
...
+spec:
+  groups:
+  - name: controller_alerts
+    rules:
+    - alert: ControllerReconciliationErrors
+      expr: |
+        sum(increase(controller_runtime_reconcile_errors_total{namespace!~"(.*-tenant|openshift-.*|kube-.*|default)"}[2m]))
+        by(container, controller, namespace, pod, service) /
+        sum(increase(controller_runtime_reconcile_total{namespace!~"(.*-tenant|openshift-.*|kube-.*|default)"}[2m]))
+        by(container, controller, namespace, pod, service) > 0.01
```
yftacherzog, line 17: Do we actually need to `sum` here? Are there multiple metrics that share the same combination of `(container, controller, namespace, pod, service)`?

hmariset: `controller_runtime_reconcile_errors_total` is unique but `controller_runtime_reconcile_total` has multiple entries with different values in result field.

test/promql/tests/test_controller_reconciliation_errors.yaml
hmariset: The series starts at 0m so at 2m the increase is 0.

prometheus/base/alerting/prometheus.controller_alerts.yaml
yftacherzog: different values that share the same combination of `(container, controller, namespace, pod, service)`? What's differentiating between those entries?

hmariset: Yes, only `result` and `value` changes for that combination for metric `controller_runtime_reconcile_total`

yftacherzog: alright. Thanks!

amisstea (approved):

PR #37, opened 2023-04-27
chore(STONEO11Y-104): Add shellcheck to the CI pipeline

Kousalya1998:
* Added shellcheck lint to Makefile.
* Added tekton task named shellscript-lint

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ RECORDING_RULE_FILES = prometheus/base/recording/*.yaml
 ALERTING_RULE_FILES = prometheus/base/alerting/*.yaml
 
 .PHONY: all
-all: prepare sync_pipenv lint test_rules pint_lint lint_yamls kustomize_build
+all: prepare sync_pipenv lint test_rules pint_lint lint_yamls kustomize_build lint_shellscript
 
 .PHONY: prepare
 prepare: pint promtool yq kustomize
@@ -69,3 +69,12 @@ kustomize_build:
 	# It will fail once we have more than one subdirectory, which will prevent us from
 	# adding untested configurations (this target will have to chage when that happens).
 	./kustomize build prometheus/* 1>/dev/null
+
+.PHONY: install_shellcheck
+install_shellcheck:
+	yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
+	yum install -y ShellCheck
+
+.PHONY: lint_shellscript
+lint_shellscript:
+	find . -type f -name "*.sh" -exec shellcheck {} + && echo "lint_shellcheck: SUCCESS"
--- a/.tekton/pull-request.yaml
+++ b/.tekton/pull-request.yaml
@@ -147,6 +147,20 @@ spec:
               image: registry.access.redhat.com/ubi9/python-39:latest
               workingDir: $(workspaces.workspace.path)
               script: make install_pipenv sync_pipenv lint_yamls
+      - name: shellscript-lint
+        workspaces:
+          - name: workspace
+            workspace: workspace
+        runAfter:
+          - clone-repository
+        taskSpec:
+          workspaces:
+            - name: workspace
+          steps:
+            - name: shellscript-lint
+              image: registry.access.redhat.com/ubi8/ubi:latest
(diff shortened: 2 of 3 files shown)


hmariset (approved):
LGTM

eisraeli:
Hi @Kousalya1998 

1. Notice you have 4 commits on the PR (while there should have been only 1).

2. Please add a short description of what is being added, in this case, ShellCheck.
For example:
"ShellCheck is an open source static analysis tool that automatically finds bugs in your shell scripts."

Makefile
```diff
@@ -69,3 +69,12 @@ kustomize_build:
 	# It will fail once we have more than one subdirectory, which will prevent us from
 	# adding untested configurations (this target will have to chage when that happens).
 	./kustomize build prometheus/* 1>/dev/null
+
+.PHONY: install_shellcheck
+install_shellcheck:
+	yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
```
hmariset, line 75: can you replace yum with dnf (just a suggestion)

Makefile
```diff
@@ -69,3 +69,12 @@ kustomize_build:
...
+	yum install -y ShellCheck
```
hmariset, line 76: same as above

Makefile
```diff
@@ -69,3 +69,12 @@ kustomize_build:
...
+
+.PHONY: lint_shellscript
+lint_shellscript:
+	find . -type f -name "*.sh" -exec shellcheck {} + && echo "lint_shellcheck: SUCCESS"
```
srivickynesh, line 80: Where is the test it's running against?

Kousalya1998: Its linting all shell scripts in a directory and its subdirectories using this command @srivickynesh

srivickynesh: There are no shell scripts all are converted to makefile.

Makefile
yftacherzog: We've been using yum for this repo. I don't think we should have yum in one place and dnf in another. If we prefer to use dnf, then I think we should have everything changed to dnf across the repo at once.

yftacherzog (changes requested):
Few points:

As @srivickynesh mentioned, this tool currently have no effect as we're not using any shell scripts within the repo, so by introducing it, we're introducing dead code to our CI, trying to foresee the future. @amisstea i see you created the ticket, can you elaborate on the rationale for doing this at this time?

Regarding the implementation, the way this PR is implemented at the moment, we're following 2 different approaches for developers and for CI: we ask developers to install it on their own (which is another hurdle for onboarding we're introducing for something that is not even used for anything), while on CI we install it automatically.

I don't think we should have it installed for developers system-wide using yum, but I think there's an alternative:

Looking at the installation guide: https://github.com/koalaman/shellcheck#installing

I see the section starts with: "Alternatively, you can download pre-compiled binaries for the latest release here".

I think this will be a more suited approach for our case, as it will allow us to:
1. Download a binary which can be run directly instead of installing it system-wide.
2. Separate the step of installing the tool from the step of using it (as it will be a binary and not part of the system config).
3. Consolidate CI and local setup in a similar way to previous tools we included.
4. Reduce the additional dependency for dev envs that is being added to the contributing guide.

Other than that, following what @eisraeli said, I think this PR is missing context and motivation for _why_ are we making those changes. It will allow the reviewers to understand what is the tool we're adding and why we're adding it. Being able to answer this before creating the PR, have the potential to also eliminate some of the questions raised here, and in turn eliminate some review cycles.

Kousalya1998:
@amisstea Alex, I need your input here.

yftacherzog:
> @amisstea Alex, I need your input here.

Alex mentioned on the stand up call that he has no additional information about why this is needed.

amisstea:
> @amisstea Alex, I need your input here.

We created this story in a refinement session soon after https://issues.redhat.com/browse/STONEO11Y-90 was complete. We probably failed to recognize this at the time. IMO we shouldn't add functionality to the CI which isn't utilized because then it's only tech debt providing no value. I'm all in favor of closing this one.

Kousalya1998:
I am closing this pull request at the moment.

PR #39, opened 2023-05-08
Add an example grafana dashboard

avi-biton:
Add an example grafana dashboard

### DIFF
--- a/grafana/dashboards/o11y-dashboard.json
+++ b/grafana/dashboards/o11y-dashboard.json
@@ -1,0 +1,127 @@
+{
+    "annotations": {
+        "list": [
+            {
+                "builtIn": 1,
+                "enable": true,
+                "hide": true,
+                "iconColor": "rgba(0, 211, 255, 1)",
+                "name": "Annotations & Alerts",
+                "target": {
+                    "limit": 100,
+                    "matchAny": false,
+                    "tags": [],
+                    "type": "dashboard"
+                },
+                "type": "dashboard"
+            }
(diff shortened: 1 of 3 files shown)


avi-biton:
@gbenhaim @amisstea This is a PR for adding an example dashboard so we can demonstrate to other teams the correct way to add dashboards

PTAL :)

grafana/dashboards/o11y-dashboard.json
amisstea, line 1: Have you considered adding some basic linting for dashboards in the CI? There's a relatively new linter that might be useful: https://github.com/grafana/dashboard-linter

avi-biton: Should we open a ticket for it? or should we add it to this one?

avi-biton: Looking at it and testing it, its in early phases and focused exclusively on dashboards that use a Prometheus data source.

We remove the datasource from our json files to get the default datasource.
In my tests it fails with error: `Dashboard 'o11y' does not have a templated data source`.
So, I think it is not suited for us.
WDYT?

amisstea: It seems we might be able to [exclude](https://github.com/grafana/dashboard-linter/blob/main/docs/index.md#exclusions-and-warnings) that [rule](https://github.com/grafana/dashboard-linter/blob/main/docs/rules/template-datasource-rule.md).

PR #40, opened 2023-05-11
chore(RHTAPWATCH-163): Add dashboard linting to o11y CI

ernesgonzalez33:
- Add Grafana dashboard linting for all dashboards in the `grafana/dashboards` directory
- Add a `.lint` file in the `grafana/dashboards` directory with the rules to be excluded in the linting

### DIFF
--- a/.tekton/pull-request.yaml
+++ b/.tekton/pull-request.yaml
@@ -148,6 +148,20 @@ spec:
               image: registry.access.redhat.com/ubi9/python-39:latest
               workingDir: $(workspaces.workspace.path)
               script: make install_pipenv sync_pipenv lint_yamls
+      - name: run-grafana-dashboard-lint
+        workspaces:
+          - name: workspace
+            workspace: workspace
+        runAfter:
+          - clone-repository
+        taskSpec:
+          workspaces:
+            - name: workspace
+          steps:
+            - name: run-grafana-dashboard-lint
+              image: registry.access.redhat.com/ubi9/go-toolset:latest
+              workingDir: $(workspaces.workspace.path)
+              script: make dashboard_linter lint_grafana_dashboards
   workspaces:
     - name: workspace
       volumeClaimTemplate:
--- a/Makefile
+++ b/Makefile
@@ -4,9 +4,11 @@ YQ_VERSION = 4.31.2
 EXTRACTED_RULE_FILE = test/promql/extracted-rules.yaml
 RECORDING_RULE_FILES = prometheus/base/recording/*.yaml
 ALERTING_RULE_FILES = prometheus/base/alerting/*.yaml
+GRAFANA_DASHBOARDS = $(shell ls grafana/dashboards) 
+GOPATH = $(shell go env GOPATH)
 
 .PHONY: all
-all: prepare sync_pipenv lint test_rules pint_lint lint_yamls kustomize_build
+all: prepare sync_pipenv lint test_rules pint_lint lint_yamls dashboard_linter lint_grafana_dashboards kustomize_build
 
 .PHONY: prepare
 prepare: pint promtool yq kustomize
@@ -46,6 +48,9 @@ yq:
 kustomize:
 	curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | bash
 
+dashboard_linter:
+	go install github.com/grafana/dashboard-linter@v0.0.0-20230531105903-cd7fcaf3bec8
+
 .PHONY: pint_lint
 pint_lint:
 	echo "Linting Prometheus rules..."
@@ -62,6 +67,13 @@ sync_pipenv:
 .PHONY: lint_yamls
 lint_yamls:
 	python3 -m pipenv run yamllint . && echo "lint_yamls: SUCCESS"
+
+.PHONY: lint_grafana_dashboards
+lint_grafana_dashboards:
+	@for dashboard in ${GRAFANA_DASHBOARDS} ; do \
+		echo -e "Linting dashboard $$dashboard\n"; \
+		${GOPATH}/bin/dashboard-linter lint grafana/dashboards/$$dashboard; done
+	@echo
 
 .PHONY: kustomize_build
 kustomize_build:
--- a/grafana/dashboards/.lint
+++ b/grafana/dashboards/.lint
@@ -1,0 +1,5 @@
+exclusions:
+  template-datasource-rule:
+    reason: "We remove the datasource from our json files to get the default datasource."
+  panel-datasource-rule:
+    reason: "We remove the datasource from our json files to get the default datasource."
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,6 +14,7 @@ This section describes the main steps required before you can add your code and 
 * Linux machine with x86-64 architecture with bash installed.
 * Basic tools: curl, tar, make.
 * Python3 and pipenv. (run: `python3 -m pip install pipenv --user`)
+* Go installed. Intructions in Go's [docs](https://go.dev/doc/install)
 
 ### Running the Automated Tests
 1. Clone your fork of the project.
@@ -25,6 +26,7 @@ The following tools are currently used for automated testing within the project:
 * [pint](https://cloudflare.github.io/pint) - PromQL linting
 * [gitlint](https://jorisroovers.com/gitlint/) - commit message linting
 * [yamllint](https://yamllint.readthedocs.io) - YAML linting
+* [dashboard-linter](https://github.com/grafana/dashboard-linter/blob/main/docs/index.md) - Grafana Dashboards linting
 
 ### Unit Tests
 PromQL unit tests are stored in YAML files under [test/promql/tests](test/promql/tests).


ernesgonzalez33:
Waiting on PR #42 for git-lint test to pass

grafana/dashboards/.lint
```diff
@@ -0,0 +1,5 @@
+exclusions:
+  template-datasource-rule:
+    reason: "We remove the datasource from our json files to get the default datasource."
+  panel-datasource-rule:
+    reason: "We remove the datasource from our json files to get the default datasource."
```
yftacherzog, line 5: add new line before end of file

.tekton/pull-request.yaml
```diff
@@ -147,6 +147,31 @@ spec:
...
+        runAfter:
+          - clone-repository
+        taskSpec:
+          workspaces:
+            - name: workspace
+          steps:
+            - name: run-grafana-dashboard-lint
+              image: registry.access.redhat.com/ubi9/go-toolset:latest
+              workingDir: $(workspaces.workspace.path)
+              script: |
```
yftacherzog, line 163: Should this test run in local dev setups as well? If it is, it should probably be implemented inside the makefile rather than inside this file.

yftacherzog: Need to update the contributing guide if there are any changes to the requirements for running the tests locally.

grafana/dashboards/.lint
ernesgonzalez33: Done!

.tekton/pull-request.yaml
ernesgonzalez33: Done both stuff in the latest commit

yftacherzog: Please make it all on a single commit. See this section: https://github.com/redhat-appstudio/o11y/blob/main/CONTRIBUTING.md#pull-request-span

ernesgonzalez33: Solved! But I think I'm triggering the dashboard-linter twice in the CI. I need to check on that

Makefile
```diff
@@ -63,6 +67,13 @@ sync_pipenv:
 lint_yamls:
 	python3 -m pipenv run yamllint . && echo "lint_yamls: SUCCESS"
 
+.PHONY: lint_grafana_dashboards
+lint_grafana_dashboards:
+	@for dashboard in ${GRAFANA_DASHBOARDS} ; do \
+		echo -e "Linting dashboard $$dashboard\n"; \
+		${HOME}/go/bin/dashboard-linter lint grafana/dashboards/$$dashboard; done
```
yftacherzog, line 74: This one assumes the installation location. It doesn't work for me:
```
Linting dashboard o11y-dashboard.json

/bin/sh: line 2: /home/yherzog/go/bin/dashboard-linter: No such file or directory
make: *** [Makefile:72: lint_grafana_dashboards] Error 127
```

It works for me on the other hand if I make this change (not sure if this is enough to assume that it will work for all):
```suggestion
		dashboard-linter lint grafana/dashboards/$$dashboard; done
```

grafana/dashboards/.lint
```diff
@@ -0,0 +1,6 @@
...
+
```
yftacherzog, line 6:
```suggestion

```

Makefile
ernesgonzalez33: I will have to research a bit on this then. Because that works for you because you have your `$GOPATH/bin` in your `$PATH`. But that's not necessarily the case for everybody. Also, using the `$GOPATH/bin/dashboard-linter` could not work when the `$GOPATH` is not set. I was too optimistic thinking that `$HOME/go/bin` would work for everybody

grafana/dashboards/.lint
ernesgonzalez33: VSCode stuff. I'll remove them

ernesgonzalez33: Done

Makefile
ernesgonzalez33: Changed. Now I set the GOPATH in the task in the CI and added to have the GOPATH env var set as one of the prerequisites in the `CONTRIBUTING` guide. I preferred doing it this way instead of setting it in the Makefile because people can have several installations of Go and it is intrusive to change that env var in the Makefile

yftacherzog: I agree on not being intrusive.

I'm starting to wonder whether the trade-off here still makes sense. We're now saying that in order to get the tests running for this repo, developers will have to either manually update their GOPATH every time or change their profile scripts. They are going to do all that to be able to test regressions on something that only very few are going to actually touch.

These are the sort of things that make people give up on running tests locally or even contributing at all. I think we should either make it more streamlined or somehow make it optional on local envs (e.g. skip the test locally if the path is not set).

@Kousalya1998 @hmariset @avi-biton @eisraeli @Omeramsc @amisstea
Can we have your thoughts/suggestions?

yftacherzog: What about something like this instead?
```
	@for dashboard in ${GRAFANA_DASHBOARDS} ; do \
		echo -e "Linting dashboard $$dashboard\n"; \
		GOPATH=$(go env GOPATH) && ${GOPATH}/bin/dashboard-linter lint grafana/dashboards/$$dashboard; done
```

Omeramsc: wouldn't that set the env var anyway? why not get the current `GOPATH`, then set it to `go env GOPATH` once before the loop, and return its original value after finishing?

yftacherzog: I edited the comment. Should be without the `&&` after setting the variable.

yftacherzog: I think we should not change it and then change back - what happens if you kill the process in the middle?

yftacherzog: I'm not sure whether my suggestion is actually working. I think it's not setting the variable as I'd expected.

avi-biton: > 	@for dashboard in ${GRAFANA_DASHBOARDS} ; do \
		echo -e "Linting dashboard $$dashboard\n"; \
		GOPATH=$(go env GOPATH) ${GOPATH}/bin/dashboard-linter lint grafana/dashboards/$$dashboard; done

I have no `GOPATH` defined in my environment, this is not defining the `GOPATH`, it is still empty

avi-biton: Do we really need the `GOPATH`?
```
@for dashboard in ${GRAFANA_DASHBOARDS} ; do \
		echo -e "Linting dashboard $$dashboard\n"; \
		dashboard-linter lint grafana/dashboards/$$dashboard; done
```

This command works in my environment, it installs the dashboard-linter in:

```
dashboard_linter:
	go install github.com/grafana/dashboard-linter@latest
```
And it does not need the `GOPATH`

yftacherzog: @ernesgonzalez33 maybe adding another variable line at the top could work. Something like:
```
...
ALERTING_RULE_FILES = prometheus/base/alerting/*.yaml
GRAFANA_DASHBOARDS = $(shell ls grafana/dashboards)
GOPATH = $(shell go env GOPATH)
```

avi-biton: > @ernesgonzalez33 maybe adding another variable line at the top could work. Something like:
> 
> ```
> ...
> ALERTING_RULE_FILES = prometheus/base/alerting/*.yaml
> GRAFANA_DASHBOARDS = $(shell ls grafana/dashboards)
> GOPATH = $(shell go env GOPATH)
> ```

Tested in my environment and it works

ernesgonzalez33: > Do we really need the `GOPATH`?
> 
> ```
> @for dashboard in ${GRAFANA_DASHBOARDS} ; do \
> 		echo -e "Linting dashboard $$dashboard\n"; \
> 		dashboard-linter lint grafana/dashboards/$$dashboard; done
> ```
> 
> This command works in my environment, it installs the dashboard-linter in:
> 
> ```
> dashboard_linter:
> 	go install github.com/grafana/dashboard-linter@latest
> ```
> 
> And it does not need the `GOPATH`

But this is because you have your `$GOPATH/bin` inside your `PATH`. That's not necessarily the case for everyone

ernesgonzalez33: > @ernesgonzalez33 maybe adding another variable line at the top could work. Something like:
> 
> ```
> ...
> ALERTING_RULE_FILES = prometheus/base/alerting/*.yaml
> GRAFANA_DASHBOARDS = $(shell ls grafana/dashboards)
> GOPATH = $(shell go env GOPATH)
> ```

I'll test this and see if it is not changing my $GOPATH. If that's the case, it is a good solution

ernesgonzalez33: Tested it and I like it. I removed the GOPATH instructions from the `CONTRIBUTING` file and also, I don't need to set up the variable in the CI anymore either

ernesgonzalez33: @yftacherzog, can I resolve this conversation and open another one in the `CONTRIBUTING.md` document for that?

yftacherzog: Sure

avi-biton:
lgtm

Omeramsc (approved):
LGTM

Makefile
```diff
@@ -46,6 +48,9 @@ yq:
 kustomize:
 	curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | bash
 
+dashboard_linter:
+	go install github.com/grafana/dashboard-linter@latest
```
gbenhaim, line 52: It would be better no to use latest. We should use a static version.

.tekton/pull-request.yaml
```diff
@@ -52,7 +52,7 @@ spec:
             - name: run-tests
               image: registry.access.redhat.com/ubi8/ubi:latest
               workingDir: $(workspaces.workspace.path)
-              script: yum install -y make && make prepare
+              script: yum install -y make go && make prepare
```
gbenhaim, line 55: do we need a specific go version?

Makefile
avi-biton: When I install the latest, it shows this prompt:
`go: downloading github.com/grafana/dashboard-linter v0.0.0-20230531105903-cd7fcaf3bec8`
We can use this version for our installation:
`go install github.com/grafana/dashboard-linter@v0.0.0-20230531105903-cd7fcaf3bec8`

ernesgonzalez33: Thanks, @avi-biton! I changed it for your way of doing it now

.tekton/pull-request.yaml
ernesgonzalez33: I removed this installation. Not needed anymore. I'm not calling the dashboard linter from the `prepare_tests`

Kousalya1998 (approved):
/lgtm

PR #42, opened 2023-05-15
fix: gitlint fails over detected dubious ownership in repository

yftacherzog:
the gitlint test is failing while fetching the target branch. This fix addresses that.


Omeramsc (approved):
LGTM

yftacherzog:
@amisstea is this something we can get merged at the current state of "frozenness". Should be harmless, and in case we need to get anything out of the door quickly, we'll have to get it merged anyhow.

yftacherzog:
> 1. We should add a JIRA issue for this fix.
> 2. The error itself as it appears in  [chore: add ernesto to owners #41](https://github.com/redhat-appstudio/o11y/pull/41) log and should be included in the issue.
> 
> ```
> + git fetch origin main
> fatal: detected dubious ownership in repository at '/workspace/workspace'
> To add an exception for this directory, call:
> 
> 	git config --global --add safe.directory /workspace/workspace
> ```
> 
> 4. The commit message should be more informative about the problem and the solution.
> 5. That being said, I'm not sure if adding this exception is the best solution.
>    I understand it will solve the error, but do we fully understand the cause?
>    IIUC, it's about files ownership on the file system.

I'd prefer not to create a Jira ticket at this case. Mainly because opening it will take me as much as it took me to create the PR. 

I changed the commit message. 

I don't know what's the reason we're seeing it now. I don't think there's a risk here (I might be wrong), so the way I see it, I think we can go with a quick fix for now and if we see any issues later, we can address them as they come.

eisraeli:
> I'd prefer not to create a Jira ticket at this case. Mainly because opening it will take me as much as it took me to create the PR.

I know. But it just creates visibility and better tracking for PRs.
 
> I changed the commit message.

Thanks
 
> I don't know what's the reason we're seeing it now. I don't think there's a risk here (I might be wrong), so the way I see it, I think we can go with a quick fix for now and if we see any issues later, we can address them as they come.

I found some more information I found about this error:
* https://github.blog/2022-04-12-git-security-vulnerability-announced/#cve-2022-24765
* https://git-scm.com/docs/git-config/2.35.2#Documentation/git-config.txt-safedirectory

yftacherzog:
> > I'd prefer not to create a Jira ticket at this case. Mainly because opening it will take me as much as it took me to create the PR.
> 
> I know. But it just creates visibility and better tracking for PRs.
> 
> > I changed the commit message.
> 
> Thanks
> 
> > I don't know what's the reason we're seeing it now. I don't think there's a risk here (I might be wrong), so the way I see it, I think we can go with a quick fix for now and if we see any issues later, we can address them as they come.
> 
> I found some more information I found about this error:
> 
> * https://github.blog/2022-04-12-git-security-vulnerability-announced/#cve-2022-24765
> * https://git-scm.com/docs/git-config/2.35.2#Documentation/git-config.txt-safedirectory

I added an issue in Jira.

My guess is that we're seeing this now because the git package on image `ubi8/ubi:latest` got updated - or something along those lines.

As I see it, adding this workaround doesn't change anything in the way we're doing things other than whatever we do on this specific gitlint task. For that reason, I don't think it introduces new risks, and that's the reason I think this is good enough for now, and doesn't worth the effort of further investigating.

eisraeli (approved):

amisstea (approved):

PR #46, opened 2023-06-12
chore(RHTAPWATCH-110): Differ between uncategorized pipelines and deployments

ShakedAviv50313:
* Differ between uncategorized pipelines and deployments


avi-biton:
LGTM, Other than this PR, we have to add another PR to add the new label `label_app_kubernetes_io_managed_by` to the remote write configuration:  
https://github.com/redhat-appstudio/infra-deployments/blob/main/components/monitoring/prometheus/base/monitoringstack/monitoringstack.yaml#L45

avi-biton (approved):

PR #47, opened 2023-06-13
chore(RHTAPWATCH-110): Add "label_app_kubernetes_io_managed_by" label to recording rule

ShakedAviv50313:
* This is a second commit regarding RHTAPWATCH-110, the label "label_app_kubernetes_io_managed_by"
was added to appstudio_container_resource_limits.

* Added test cases.

### DIFF
--- a/prometheus/base/recording/prometheus.rules.yaml
+++ b/prometheus/base/recording/prometheus.rules.yaml
@@ -21,7 +21,7 @@ spec:
           expr: |
             {__name__=~"kube_pod_.*container_resource_limits"}
             * on (namespace, pod)
-            group_left(label_pipelines_appstudio_openshift_io_type) kube_pod_labels
+            group_left(label_pipelines_appstudio_openshift_io_type, label_app_kubernetes_io_managed_by) kube_pod_labels
     - name: appstudio-resource-gauge
       interval: 1m
       rules:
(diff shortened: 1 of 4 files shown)


yftacherzog (changes requested):
The commit message title and PR title assume some sort of context, which might make sense now, but might make less sense if someone is to look at this change somewhere in the future.

Also, the commit message title should usually be written in an imperative form (i.e. as if you're commanding something). In this case, it means using `add label...` rather than `adding label...`.

I suggest changing the PR title and commit message title to something like this instead:

```
Add label to prometheus recording rule
```
You can mention the test cases, but I don't think this is necessary. It's expected that when we add a feature or fix a bug, we also add tests for that.

ShakedAviv50313:
@yftacherzog  Done:)

yftacherzog (approved):

yftacherzog:
> @yftacherzog Done:)

gitlint fails now. There's a limitation on the number of characters for the commit title. This cannot be addressed by breaking the title line, as the line below it needs to be empty.

It implies the need for being brief when writing the title. We should drop stuff that are unnecessary for the the title, and we can add them to the body (in this case, the name of the label can be dropped from the title).

Omeramsc (approved):
Besides Yftach's comments, LGTM.

avi-biton (approved):

PR #49, opened 2023-06-21
fix(RHTAPWATCH-153): Prepare alerting rules for RHOBS

Kousalya1998:
update alerting rules according to RHOBS

 Signed-off-by: Kousalya Lakshmanan <klakshma@redhat.com>

### DIFF
--- a/rhobs/alerting/prometheus.controller_alerts.yaml
+++ b/rhobs/alerting/prometheus.controller_alerts.yaml
@@ -1,13 +1,13 @@
 apiVersion: monitoring.coreos.com/v1
 kind: PrometheusRule
 metadata:
+  name: rhtap-controller-alerting-rules
   labels:
-    openshift.io/prometheus-rule-evaluation-scope: leaf-prometheus
-  name: o11y-controller-alerting-rules
-  namespace: o11y
+    tenant: rhtap
 spec:
   groups:
   - name: controller_alerts
+    interval: 1m
     rules:
     - alert: ControllerReconciliationErrors
       expr: |
--- a/rhobs/alerting/prometheus.pod_alerts.yaml
+++ b/rhobs/alerting/prometheus.pod_alerts.yaml
@@ -1,13 +1,13 @@
 apiVersion: monitoring.coreos.com/v1
 kind: PrometheusRule
 metadata:
+  name: rhtap-pod-alerting-rules
   labels:
-    openshift.io/prometheus-rule-evaluation-scope: leaf-prometheus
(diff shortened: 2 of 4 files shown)


amisstea:
Please amend the commit message to make it clear that this is more like prep work for the deployment rather than actually enabling deployment of these rules.

Shouldn't we also make sure these rules are no longer deployed to the RHTAP dataplane clusters?

yftacherzog:
> Please amend the commit message to make it clear that this is more like prep work for the deployment rather than actually enabling deployment of these rules.
> 
> Shouldn't we also make sure these rules are no longer deployed to the RHTAP dataplane clusters?

@amisstea we should. I wonder how we should go about it. Need to keep in mind:

1. We don't make changes that will have adverse affect on what is being deployed with argoCD.
2. We keep the path of the alerting rules in a form that can be consumed by app-interface (e.g. contains only rules and just the relevant ones).

Preferably, I want us to make those changes without making too many changes at the same time.

I think we first need to make it so that we're not deploying anything with argoCD, and then we can make changes more freely.

I think the reality is that infra-deployments has no reasons at the moment to deploy anything coming from the o11y repo. Would you agree? Should we offboard o11y from infra-deployments?

avi-biton:
> I think the reality is that infra-deployments has no reasons at the moment to deploy anything coming from the o11y repo. Would you agree? Should we offboard o11y from infra-deployments?

It seems that `o11y` only supplies recording and alert rules to `infra-deployments` (which should not be there), I am all for removing it from `infra-deployments`

yftacherzog:
> > I think the reality is that infra-deployments has no reasons at the moment to deploy anything coming from the o11y repo. Would you agree? Should we offboard o11y from infra-deployments?
> 
> It seems that `o11y` only supplies recording and alert rules to `infra-deployments` (which should not be there), I am all for removing it from `infra-deployments`

@avi-biton were looking into that, and saw that the o11y namespace is still in use (e.g. for the dummy service monitor example). For now, I created a PR in this repo that will offboard the deployment of the rules specifically.

https://github.com/redhat-appstudio/o11y/pull/50

yftacherzog:
@amisstea can we delete the argoCD-related rules?

yftacherzog:
@Kousalya1998 this PR is now the last step before we can have app-interface reference the rules in this repo. Can you rebase the PR and make sure all rules are up to date in the new location (`rhobs/alerting`)?

yftacherzog (approved):

amisstea (approved):

PR #51, opened 2023-06-22
chore(RHTAPWATCH-297): Modify the alerting rules and test cases with source_cluster information

Kousalya1998:
Add source cluster information to alerting rules and description
Add source cluster details to the test cases


amisstea (approved):

yftacherzog (approved):
lgtm. I think it can be merged already. Don't think we have any blockers for this one

PR #52, opened 2023-06-25
refactor(RHTAPWATCH-330): delete recording rules and move alert rules

yftacherzog:
We no longer need to deploy the recording rules. On top of that, we should only deploy the alerting rules for data plane monitoring. Those rules should be moved to a different path as they are to be deployed to RHOBS by app-interface.

This change removes the unneeded rules and changes the path for the rest.


yftacherzog:
@Kousalya1998 @avi-biton @amisstea please take a look

amisstea (approved):

Kousalya1998 (approved):
lgtm

avi-biton (commented):
Don't we need the argocd alerts?

yftacherzog:
> Don't we need the argocd alerts?

app-interface is going to deploy the data plane alerting rules to RHOBS by referencing the location here.
It's already deploying the control plane rules (those alerts we already see in Slack) as defined here: https://gitlab.cee.redhat.com/service/app-interface/-/tree/master/resources/stonesoup/argocd-control-plane/monitoring

We should have each of the rules maintained only at the location from which it's deployed.

avi-biton (approved):
LGTM

PR #68, opened 2023-08-17
chore(RHTAPWATCH-334): Remove the o11y namespace resource definition from the o11y repo

Kousalya1998:
* The o11y namespace definition and all kustomization files should be removed from the o11y repo
 * The [prometheus/base path](https://github.com/redhat-appstudio/o11y/tree/main/prometheus) should be removed and the kustomize build test should be removed from the makefile and the [CI file](https://github.com/redhat-appstudio/o11y/blob/main/.tekton/pull-request.yaml).
 * kustomize installation should be removed from local tests and CI.


amisstea (approved):

Omeramsc (approved):
LGTM

PR #73, opened 2023-09-11
chore(RHTAPWATCH-360): update o11y README file

Kousalya1998:
- Add alerting rules with description
- Remove Grafana dashboard details from README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,28 @@ Control plane clusters alert rules are maintained by the same team, but are kept
 different
 [repository](https://gitlab.cee.redhat.com/service/app-interface/-/tree/master/resources/stonesoup/argocd-control-plane/monitoring).
 
+The different alerting rules in this repository are:
+
+## Control Plane Alerts
+
+* [**Alert Rule ProgressingArgocdApp**](https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/o11y/alert-rule-ProgressingArgocdApp.md)
+
+* [**Alert Rule DegradedArgocdApp**](https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/o11y/alert-rule-degradedArgocdApp.md)
+
+## Data Plane Alerts
+
+* [**Alert Rule ControllerReconciliationErrors**](https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/o11y/alert-rule-controllerReconciliationErrors.md?ref_type=heads)
+
+* [**Alert Rule Unschedulable**](https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/o11y/alert-rule-unschedualablePods.md)
+
+* [**Alert Rule CrashLoopBackOff**](https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/o11y/alert-rule-crashLoopBackOff.md?ref_type=heads)
+
+* [**Alert Rule PodsNotReady**](https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/o11y/alert-rule-PodsNotReady.md?ref_type=heads)
+
+* [**Alert Rule PersistentVolumeIssues**](https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/o11y/alert-rule-pesistentVolumeIssues.md?ref_type=heads)
+
+* [**Alert Rule QuotaExceeded**](https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/o11y/alert-rule-QuotaExceeded.md)
+
 ### Updating Data Plane Alerts
 
 Alert rules for data plane clusters are being deployed by app-interface to RHOBS, to where the data plane metrics are also being forwarded. For deploying the alert rules,


yftacherzog (changes requested):
The Grafana stuff here is fairly new. This is not the same Grafana documentation that was here when this ticket was created. I don't think it should be removed. Please confirm with @amisstea.

README.md
```diff
@@ -13,6 +12,33 @@ Control plane clusters alert rules are maintained by the same team, but are kept
 different
 [repository](https://gitlab.cee.redhat.com/service/app-interface/-/tree/master/resources/stonesoup/argocd-control-plane/monitoring).
 
+The different alerting rules in this repository are:
+
+* **Alert Rule ControllerReconciliationErrors**
+
+    The alert indicates that a controller's reconciliation failure ratio is higher than the
+    allowed threshold. Controller reconciliation is the process of synchronizing resources over which a controller is watching with the business logic specifications the controller implements. Controller reconciliation errors happen when a controller fails enforcing a desired state (the specifications) over a resource it's watching. This alert is triggered only for controllers under RHTAP service namespaces. Namely, it will not be triggered for neither controllers used for OpenShift's built-in functionality nor for controllers in tenant namespaces.
```
yftacherzog, line 20: Let's keep the lines of text at a fixed length (same for the rest of the document):
```suggestion
    The alert indicates that a controller's reconciliation failure ratio is higher than
    the allowed threshold. Controller reconciliation is the process of synchronizing
    resources over which a controller is watching with the business logic specifications
    the controller implements. Controller reconciliation errors happen when a controller
    fails enforcing a desired state (the specifications) over a resource it's watching.
    This alert is triggered only for controllers under RHTAP service namespaces. Namely,
    it will not be triggered for neither controllers used for OpenShift's built-in
    functionality nor for controllers in tenant namespaces.
```

yftacherzog:
@amisstea since this ticket was created, we introduced the alert SOPs, which contains a superset of the descriptions provided here. Do you think we can link to the SOPs here instead of copying stuff from there? The only reason I see not to do it is that the SOPs are for the downstream product and they are hosted on our gitlab. Is that a good-enough reason to maintain those descriptions in two different places?

README.md
```diff
@@ -33,13 +59,6 @@ reference in app-interface to the commit hash of the changes you made.
 3. Once merged and ready to be promoted to production, update the
 [production environment](https://gitlab.cee.redhat.com/service/app-interface/-/blob/0486ef164e70259e5b85c46ab749529238368414/data/services/osd-operators/cicd/saas/saas-rhtap-rules.yaml#L39) reference in a similar manner.
 
-## Grafana Dashboards
-
-Refer to the app-interface [instructions](
-https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/monitoring.md#visualization-with-grafana)
-to learn how to develop AppSRE dashboards for RHTAP. This repository serves as
-versioned storage for the [dashboard definitions](dashboards/) and nothing more.
-
```
amisstea, line 42: These grafana dashboards are quite new and should not be removed.

amisstea:
> @amisstea since this ticket was created, we introduced the alert SOPs, which contain a superset of the descriptions provided here. Do you think we can link to the SOPs here instead of copying stuff from there? The only reason I see not to do it is that the SOPs are for the downstream product and they are hosted on our gitlab. Is that a good-enough reason to maintain those descriptions in two different places?

I agree. Let's not repeat these descriptions which will inevitably stray from the SOPs. Providing a link to gitlab is sufficient even if it's not accessible publicly.

amisstea (changes requested):

README.md
Kousalya1998: @amisstea & @yftacherzog Thanks for your inputs. I have added the Grafana dashboards back and have also added GitLab links for the alerting rules.

README.md
```diff
@@ -13,6 +13,18 @@ Control plane clusters alert rules are maintained by the same team, but are kept
 different
 [repository](https://gitlab.cee.redhat.com/service/app-interface/-/tree/master/resources/stonesoup/argocd-control-plane/monitoring).
 
+The different alerting rules in this repository are:
```
yftacherzog, line 16: What's the reason to have the links pointing to a specific commit? It means that if an SOP changes, we'll need to update those links.

Kousalya1998: @yftacherzog  I didn't notice it. I have updated it to the main branch now.

README.md
```diff
@@ -13,6 +13,18 @@ Control plane clusters alert rules are maintained by the same team, but are kept
...
+
+* [**Alert Rule ControllerReconciliationErrors**](https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/o11y/alert-rule-controllerReconciliationErrors.md?ref_type=heads)
+
+* [**Alert Rule CrashLoopBackOff**](https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/o11y/alert-rule-crashLoopBackOff.md?ref_type=heads)
+
+* [**Alert Rule PodsNotReady**](https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/o11y/alert-rule-PodsNotReady.md?ref_type=heads)
+
+* [**Alert rule for PersistentVolumeIssues**](https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/o11y/alert-rule-pesistentVolumeIssues.md?ref_type=heads)
+
+* [**Alert Rule QuotaExceeded**]()
```
yftacherzog, line 26: There's no link here

README.md
```diff
@@ -13,6 +13,18 @@ Control plane clusters alert rules are maintained by the same team, but are kept
...
 
+The different alerting rules in this repository are:
+
+* [**Alert Rule ControllerReconciliationErrors**](https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/o11y/alert-rule-controllerReconciliationErrors.md?ref_type=heads)
+
+* [**Alert Rule CrashLoopBackOff**](https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/o11y/alert-rule-crashLoopBackOff.md?ref_type=heads)
+
+* [**Alert Rule PodsNotReady**](https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/o11y/alert-rule-PodsNotReady.md?ref_type=heads)
+
+* [**Alert rule for PersistentVolumeIssues**](https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/o11y/alert-rule-pesistentVolumeIssues.md?ref_type=heads)
```
yftacherzog, line 24: Let's make the format and the capitalization consistent over all lines.

README.md
Kousalya1998: @yftacherzog  The alerts for the control plane and data plane have just been updated. I couldn't find a link to that one initially, but I have updated it.

README.md
Kousalya1998: @yftacherzog Where exactly are you referring to the change to Capitalization?

yftacherzog: rule -> Rule

yftacherzog (approved):

amisstea (approved):

PR #74, opened 2023-09-12
chore(RHTAPWATCH-472): Copy alerts from Control Plane

avi-biton:
- Create different folders for data-plane and control-plane
- Copy the control-plane alerts from control-plane to o11y
- Update both alerts and tests to comply with the testing framework
- Update the Makefile


amisstea (approved):

yftacherzog (approved):
lgtm

PR #78, opened 2023-09-18
fix(RHTAPWATCH-523): control plane alerts unsupported severity

yftacherzog:
The control plane alerts contained severity `medium` which is not one of the values supported by PagerDuty, which caused errors in RHOBS. This change addresses that.


avi-biton (approved):
/lgtm

PR #84, opened 2023-10-03
chore(RHTAPWATCH-404): update probe

avi-biton:
Update probe to select only from the latest grafana api 
Fix lint issues

### DIFF
--- a/config/probes/monitoring/grafana/base/grafana-datasource-probe.yaml
+++ b/config/probes/monitoring/grafana/base/grafana-datasource-probe.yaml
@@ -48,15 +48,12 @@ spec:
           restartPolicy: Never
           containers:
           - name: datasource-availability
-            image: registry.redhat.io/openshift4/ose-cli:latest
+            image: registry.redhat.io/openshift4/ose-cli:v4.12.0
             command: 
             - /bin/bash
             - -c
             - |
-              cd $(mktemp -d)
-              wget -O jq https://github.com/jqlang/jq/releases/download/jq-1.7/jq-linux-i386 && chmod +x jq
-              echo 'Start test'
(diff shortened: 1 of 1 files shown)


config/probes/monitoring/grafana/base/grafana-datasource-probe.yaml
```diff
@@ -48,19 +48,27 @@ spec:
...
+              requests:
+                cpu: 100m
+                memory: 10Mi
+              limits:
+                cpu: 100m
+                memory: 100Mi
+            securityContext:
+              readOnlyRootFilesystem: true
+              runAsNonRoot: true
+
```
yftacherzog, line 74: Newline at end of file

config/probes/monitoring/grafana/base/grafana-datasource-probe.yaml
```diff
@@ -11,7 +11,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: read-grafana
 rules:
-- apiGroups: ['grafana.integreatly.org']
+- apiGroups: ['grafanas.grafana.integreatly.org']
```
hugares, line 14: Should't the api group be `grafana.integreatly.org` and the resource is `grafanas`

yftacherzog (commented):
Would it fail the job if grep comes empty?

config/probes/monitoring/grafana/base/grafana-datasource-probe.yaml
avi-biton: Yes, it is leftover from my tries :)

avi-biton:
> Would it fail the job if grep comes empty?

Yes, I tested it

yftacherzog (approved):

config/probes/monitoring/grafana/base/grafana-datasource-probe.yaml
hugares: looks like I do not have the rights to resolve the comment thread I created... You will have to do it for me :)

PR #86, opened 2023-10-05
chore(STONEINTG-558): add alerting rule for IS response

jencull:


rhobs/alerting/data_plane/integration_service_response.yaml
```diff
@@ -0,0 +1,24 @@
...
+  labels:
+    tenant: rhtap
+spec:
+  groups:
+  - name: integration-service-response-exceeded
+    interval: 1m
+    rules:
+    - alert: IntegrationServiceResponse
+      expr:  sum(increase(integration_svc_response_seconds_bucket[$__interval])) by (le)
+
```
MartinBasti, line 14: this empty line shouldn't be here imho, and also at the end of the should be a newline separator

MartinBasti:
Test failed
```
time="2023-10-05T15:49:22Z" level=info msg="checking rules..."
time="2023-10-05T15:49:22Z" level=fatal msg="[rhobs/alerting/data_plane/integration_service_response.yaml] 'metadata.name' attribute does not starts with 'rhtap-' (value is 'integration-service-response-alerting-rules')\n"
make: *** [Makefile:23: check-and-test] Error 1
```

dirgim (approved):
LGTM

test/promql/tests/data_plane/pipeline_to_snapshot_test.yaml
```diff
@@ -0,0 +1,108 @@
...
+
+tests:
+
+ # Scenario when alert should NOT trigger
+  - interval: 1m
+    input_series:
+      - series: 'integration_svc_response_seconds_bucket{le="30"}'
+        values: '0+95x1'  # 95 occurrances took less than 30s
+      - series: 'integration_svc_response_seconds_bucket{le="+Inf"}'
+        values: '0+100x1'  # 100 total occurrences
```
yftacherzog, line 14: This test case is providing only 2-min-worth of measurements while the rule is applied on a time frame of 5 minutes + 1 minute of waiting. We need some more samples  and later eval_time for this one, or to have another test case with those.

test/promql/tests/data_plane/pipeline_to_snapshot_test.yaml
```diff
@@ -0,0 +1,108 @@
...
+              severity: warning
+              slo: true
+            exp_annotations:
+              summary: >-
+                PipelineRunFinish to SnapshotInProgress time exceeded
+              description: >-
+                Time from pipeline run finished to snapshot marked in progress has been over 30s for more than 10% of requests during the last 5 minutes
+              runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/integration-service/pipeline_to_snapshot_exceeded.md
+
+ # Scenario when the alert should trigger at 10.1%
```
yftacherzog, line 41: This scenario, and also the next one, contains many metrics which are not covered by the rule. I don't think we should have those metrics.

On the other hand, the rule uses the `sum` function, but we're not testing that.

looking at this metric in staging and [production](https://grafana.app-sre.devshift.net/explore?orgId=1&left=%7B%22datasource%22:%22P22466E8E7855F1E0%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22P22466E8E7855F1E0%22%7D,%22editorMode%22:%22code%22,%22expr%22:%22integration_svc_response_seconds_bucket%7Ble%3D%5C%2230%5C%22%7D%22,%22legendFormat%22:%22__auto%22,%22range%22:true,%22instant%22:true%7D%5D,%22range%22:%7B%22from%22:%22now-1h%22,%22to%22:%22now%22%7D%7D), I see that the value for `integration_svc_response_seconds_bucket{le="30"}` is generated once per cluster.

It means that this rule effectively sums over the clusters for a given environment (i.e. alerts based on the stage average and over the prod average). If that is the intention, then we should have that tested. If not, then I don't think we should use sum, or at least not in this way.

yftacherzog (changes requested):

test/promql/tests/data_plane/pipeline_to_snapshot_test.yaml
jencull: I have written new tests that cover edge % and test across 2 clusters and removed the other buckets/inputs. Apologies if I have misunderstood your recommendations. Happy to write additional tests or change the new ones if needs be.

yftacherzog: I still think that the use of sum on the alert itself might not be what you're after.

A couple of questions:

1. What's the original reason for using sum here?
2. Suppose that in one cluster, all requests take forever, and on the other they all take less than 30s, I suspect we'd like to have the alert triggered. With the current implementation, this will depend on the number of requests done on each cluster (as they will be averaged).

I think the thing to do now is to understand what should be the expected behavior.

jencull: rewritten to test with 2 clusters and different scenarios: test to verify both clusters within range, one of two out of range and two of two out of range.

yftacherzog: @jencull can you confirm that summing between the clusters is the expected behavior?

jencull: @yftacherzog, each cluster will be tested individually but the metrics are aggregated then for display on grafana

rhobs/alerting/data_plane/prometheus.pipeline_to_snapshot_alerts.yaml
```diff
@@ -0,0 +1,30 @@
...
+spec:
+  groups:
+  - name: pipeline-to-snapshot-exceeded
+    interval: 1m
+    rules:
+    - alert: PipelineRunToSnapshotExceeded
+      expr: |
+       (
+        (
+         sum(increase(integration_svc_response_seconds_bucket{le="+Inf"}[5m])) by (namespace, source_cluster) -
```
yftacherzog, line 16: Is there any difference between the result of `increase(...)` and the result of `sum(increase(...) by (...)`.
I think the current implementation is logically ok, but I don't see the added value with summing the result of `increase`.

jencull: > Is there any difference between the result of `increase(...)` and the result of `sum(increase(...) by (...)`. I think the current implementation is logically ok, but I don't see the added value with summing the result of `increase`.

I tried that initially and it won't work without 'sum'. I have tried again there now and it just won't calculate.

yftacherzog: `has_application_failed_deletion_total` in `application_alerts` file has more than one time series per cluster, so there it might be that summing makes sense.

the `by ...` part affects how the time series returned by the query inside the `sum` call will be aggregated. Currently, the label combination we give there is already unique between all the results so it effectively doesn't sum anything.

The reason it won't work for you without `sum` is that the arithmetic operators (`-`, `/`) you use cannot match between the time series that you provide as inputs.

Prometheus will only run operations between time series with identical labels. In our case, we try to subtract and divide between time series that differ between each other by the `le` label, so we need to tell Prometheus either exactly which labels it should try to match or the labels that it should ignore.

Something like this:
```
       (
        (
         increase(integration_svc_response_seconds_bucket{le="+Inf"}[5m])
         - ignoring(le)
         increase(integration_svc_response_seconds_bucket{le="30"}[5m])
        ) / ignoring(le)
        increase(integration_svc_response_seconds_bucket{le="+Inf"}[5m])
       ) > 0.10
```

rhobs/alerting/data_plane/prometheus.pipeline_to_snapshot_alerts.yaml
```diff
@@ -0,0 +1,30 @@
...
+       ) > 0.10
+      for: 1m
+      labels:
+        severity: warning
+        slo: true
+      annotations:
+        summary: >-
+          PipelineRunFinish to SnapshotInProgress time exceeded
+        description: >-
+          Time from pipeline run finished to snapshot marked in progress has been over 30s for more than 10% of requests during the last 5 minutes on cluster {{ $labels.source_cluster }}
```
yftacherzog, line 29: Let's keep the line length less than 90 chars. You should be able to do the same on the expected results, if you start there with `>-` as well.
```suggestion
        description: >-
          Time from pipeline run finished to snapshot marked in progress has been over
          30s for more than 10% of requests during the last 5 minutes on cluster
          {{ $labels.source_cluster }}
```

jencull: done.

rhobs/alerting/data_plane/prometheus.pipeline_to_snapshot_alerts.yaml
jencull: thank you so much for your help and patience.

yftacherzog (approved):

PR #89, opened 2023-10-13
chore(SVPI-645):  Add visualization for 5xx errors rate from SPs SLO

mshaposhnik:
This PR adds a visualization of the 5xx errors rate from SPs  to the RHTAP SLO  Grafana dashboard.
This is how it looks on my test dashb:
зображення

### DIFF
--- a/dashboards/grafana-dashboard-rhtap-slos.configmap.yaml
+++ b/dashboards/grafana-dashboard-rhtap-slos.configmap.yaml
@@ -197,6 +197,19 @@ data:
             "x": 0,
             "y": 11
           },
+          "id": 40,
+          "panels": [],
+          "title": "5xx errors SLO",
+          "type": "row"
+        },
+        {
+          "collapsed": false,
+          "gridPos": {
+            "h": 1,
+            "w": 24,
+            "x": 0,
+            "y": 20
+          },
           "id": 33,
           "panels": [],
           "title": "SLO Template",
@@ -212,7 +225,7 @@ data:
             "h": 7,
             "w": 10,
             "x": 0,
-            "y": 12
+            "y": 21
           },
           "id": 10,
           "links": [],
@@ -281,7 +294,7 @@ data:
             "h": 7,
             "w": 5,
             "x": 10,
-            "y": 12
+            "y": 21
           },
           "id": 6,
           "links": [],
@@ -356,7 +369,7 @@ data:
             "h": 7,
             "w": 5,
             "x": 15,
-            "y": 12
+            "y": 21
           },
           "id": 29,
           "links": [],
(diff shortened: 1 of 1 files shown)


avi-biton:
I am not sure if we want all the SLO dashboards to be in the same file.
1 - It will be a long file and hard to maintain.
2 - An error or missconfiguration of one dashboard might effect the other dashboards.

mshaposhnik:
> I am not sure if we want all the SLO dashboards to be in the same file.
@avi-biton who should make a decision on that?

avi-biton:
> > I am not sure if we want all the SLO dashboards to be in the same file.
> 
> @avi-biton who should make a decision on that?

I am adding @amisstea

amisstea:
> > > I am not sure if we want all the SLO dashboards to be in the same file.
> > 
> > 
> > @avi-biton who should make a decision on that?
> 
> I am adding @amisstea

I think we should stick to a single dashboard for now. One of the SRE onboarding requirements is to have a "[single pane of glass](https://gitlab.cee.redhat.com/service/sres-acceptance-criteria/-/blob/main/acs/OBSERVABILITY-0004-grafana-dashboard-for-slis-and-slos.md)" to view SLOs. Distributing them across multiple dashboards would go against this guidance meaning we'd need to seek approval from AppSRE. That doesn't preclude us from having additional dashboards in case more detail is required.

dashboards/grafana-dashboard-rhtap-slos.configmap.yaml
```diff
@@ -387,6 +387,126 @@ data:
...
+              "displayMode": "list",
+              "placement": "bottom",
+              "showLegend": true
+            },
+            "tooltip": {
+              "mode": "single",
+              "sort": "none"
+            }
+          },
+          "targets": [
```
amisstea, line 479: Can we combine all these targets to produce a single metric? An SLO should produce a single numerical value. AppSRE also requires a single `expr` in the SLO doc schema.

avi-biton:
> > > > I am not sure if we want all the SLO dashboards to be in the same file.
> > > 
> > > 
> > > @avi-biton who should make a decision on that?
> > 
> > 
> > I am adding @amisstea
> 
> I think we should stick to a single dashboard for now. One of the SRE onboarding requirements is to have a "[single pane of glass](https://gitlab.cee.redhat.com/service/sres-acceptance-criteria/-/blob/main/acs/OBSERVABILITY-0004-grafana-dashboard-for-slis-and-slos.md)" to view SLOs. Distributing them across multiple dashboards would go against this guidance meaning we'd need to seek approval from AppSRE. That doesn't preclude us from having additional dashboards in case more detail is required.

Thanks for the explanation @amisstea

dashboards/grafana-dashboard-rhtap-slos.configmap.yaml
mshaposhnik: Done. Now values are combined for all SP-s.

dashboards/grafana-dashboard-rhtap-slos.configmap.yaml
```diff
@@ -387,6 +400,108 @@ data:
...
+            {
+              "datasource": "$datasource",
+              "editorMode": "builder",
+              "expr": "sum(increase(redhat_appstudio_spi_service_provider_response_time_seconds_count{status=~\"5..\"}[24h]) OR on() vector(0)) / sum(increase(redhat_appstudio_spi_service_provider_response_time_seconds_count{}[24h]))",
```
amisstea, line 496: Instead of using `24h`, I think we should use `$__range` instead. The dashboard uses a default time window of '28d' which matches the sliding window over which the SLO will be evaluated.

mshaposhnik:
Ok i gave up on this, since new changes are merged into main, i have to start everything from scratch.

PR #94, opened 2023-10-19
feat(GITOPSRVCE-750): Add alert rule/test for Synced Applications SLO

keithchong:
See https://issues.redhat.com/browse/GITOPSRVCE-750.  Alert rule and test.

### DIFF
--- a/rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
@@ -12,13 +12,13 @@ spec:
     - alert: GitOpsApplicationSyncErrors
       expr: |
         (
-          sum(avg_over_time(argocd_app_info{sync_status="Synced", namespace="gitops-service-argocd"}[5m]))
+          sum(avg_over_time(argocd_app_info{sync_status="Synced", namespace="gitops-service-argocd", source_cluster="stonesoupp01ue1"}[5m]))
           or
           sum(argocd_app_info{namespace="gitops-service-argocd"}) * 0
         )
         /
         (
-          sum(avg_over_time(argocd_app_info{namespace="gitops-service-argocd"}[5m]))
+          sum(avg_over_time(argocd_app_info{namespace="gitops-service-argocd", source_cluster="stonesoupp01ue1" }[5m]))
         ) < 0.95
       for: 5m
       labels:
--- a/rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
@@ -11,16 +11,24 @@ spec:
     rules:
     - alert: GitOpsApplicationSyncErrors
       expr: |
-        sum(argocd_app_info{sync_status="Synced", namespace="gitops-service-argocd"})
+        (
+          sum(avg_over_time(argocd_app_info{sync_status="Synced", namespace="gitops-service-argocd"}[5m]))
+          or
+          sum(argocd_app_info{namespace="gitops-service-argocd"}) * 0
+        )
         /
-        sum(argocd_app_info{namespace="gitops-service-argocd"}) < 0.95
+        (
+          sum(avg_over_time(argocd_app_info{namespace="gitops-service-argocd"}[5m]))
+          or
+          sum(argocd_app_info{namespace="gitops-service-argocd"}) * 0
+        ) < 0.95
       for: 5m
       labels:
         severity: warning
         slo: true
       annotations:
         summary: >-
-          Less than 95% of GitOps applications are in synced state
+          Less than 95% of GitOps applications are in synced state: {{ $value }}
         description: >-
           The percentage total of all Argo CD applications that are in Synced state is less than 95%.
         runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/gitops/deploy-from-git-to-k8s.md
--- a/rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
@@ -12,13 +12,14 @@ spec:
     - alert: GitOpsApplicationSyncErrors
       expr: |
         (
-          sum(avg_over_time(argocd_app_info{sync_status="Synced", namespace="gitops-service-argocd", source_cluster="stonesoupp01ue1"}[5m]))
+          sum(avg_over_time(argocd_app_info{sync_status="Synced", namespace="gitops-service-argocd"}[5m])) by (source_cluster)
           or
           sum(argocd_app_info{namespace="gitops-service-argocd"}) * 0
         )
         /
         (
-          sum(avg_over_time(argocd_app_info{namespace="gitops-service-argocd", source_cluster="stonesoupp01ue1" }[5m]))
+          sum(avg_over_time(argocd_app_info{namespace="gitops-service-argocd"}[5m]))
+          by (source_cluster)
         ) < 0.95
       for: 5m
       labels:
--- a/rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
@@ -12,13 +12,14 @@ spec:
     - alert: GitOpsApplicationSyncErrors
       expr: |
         (
-          sum(avg_over_time(argocd_app_info{sync_status="Synced", namespace="gitops-service-argocd"}[5m])) by (source_cluster)
+          sum(last_over_time(argocd_app_info{sync_status="Synced", namespace="gitops-service-argocd"}[1m]))
+          by (source_cluster)
           or
-          sum(argocd_app_info{namespace="gitops-service-argocd"}) * 0
+          sum(argocd_app_info{namespace="gitops-service-argocd"}) by (source_cluster) * 0
         )
         /
         (
-          sum(avg_over_time(argocd_app_info{namespace="gitops-service-argocd"}[5m]))
+          sum(last_over_time(argocd_app_info{namespace="gitops-service-argocd"}[1m]))
           by (source_cluster)
         ) < 0.95
       for: 5m
--- a/rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
@@ -19,8 +19,6 @@ spec:
         /
         (
           sum(avg_over_time(argocd_app_info{namespace="gitops-service-argocd"}[5m]))
-          or
-          sum(argocd_app_info{namespace="gitops-service-argocd"}) * 0
         ) < 0.95
       for: 5m
       labels:
@@ -28,25 +26,7 @@ spec:
         slo: true
       annotations:
         summary: >-
-          Less than 95% of GitOps applications are in synced state: {{ $value }}
+          Less than 95% of GitOps applications are in synced state: {{ $value | humanizePercentage }}
         description: >-
           The percentage total of all Argo CD applications that are in Synced state is less than 95%.
-        runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/gitops/deploy-from-git-to-k8s.md
-    - alert: GitOpsApplicationNoSyncApps
-      expr: |
-        (
-          sum(argocd_app_info{sync_status="OutOfSync", namespace="gitops-service-argocd"}) +
-          sum(argocd_app_info{sync_status="Unknown", namespace="gitops-service-argocd"})
-        )
-        /
-        sum(argocd_app_info{namespace="gitops-service-argocd"}) >= 1
-      for: 5m
-      labels:
-        severity: warning
-        slo: true
-      annotations:
-        summary: >-
-          There are no apps in Synced state but there are one or more apps in OutOfSync or Unknown states
-        description: >-
-          There are no apps in Synced state but there are one or more apps in OutOfSync or Unknown states
         runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/gitops/deploy-from-git-to-k8s.md
--- a/rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
@@ -1,0 +1,28 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: rhtap-gitops-sync-alerting
+  labels:
+    tenant: rhtap
+spec:
+  groups:
+  - name: gitops_sync_alerts
+    interval: 1m
+    rules:
+    - alert: GitOpsApplicationSyncErrors
+      expr: |
+        sum(last_over_time(argocd_app_info{sync_status="Synced"}[5m]))
+        /
+        (
+          sum(last_over_time(argocd_app_info{sync_status="Synced"}[5m])) + sum(last_over_time(argocd_app_info{sync_status="OutOfSync"}[5m])) + sum(last_over_time(argocd_app_info{sync_status="Unknown"}[5m]))
+        ) < 0.95
+      for: 5m
+      labels:
+        severity: warning
+        slo: true
+      annotations:
+        summary: >-
+          Less than 95% of GitOps applications are in synced state
+        description: >-
+          The percentage total of all Argo CD applications that are in Synced state is less than 95%.
+        runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/gitops/deploy-from-git-to-k8s.md
--- a/rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
@@ -24,3 +24,21 @@ spec:
         description: >-
           The percentage total of all Argo CD applications that are in Synced state is less than 95%.
         runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/gitops/deploy-from-git-to-k8s.md
+    - alert: GitOpsApplicationNoSyncApps
+      expr: |
+        (
+          sum(argocd_app_info{sync_status="OutOfSync", namespace="gitops-service-argocd"}) +
+          sum(argocd_app_info{sync_status="Unknown", namespace="gitops-service-argocd"})
+        )
+        /
+        sum(argocd_app_info{namespace="gitops-service-argocd"}) >= 1
+      for: 5m
+      labels:
+        severity: warning
+        slo: true
+      annotations:
+        summary: >-
+          There are no apps in Synced state but there are one or more apps in OutOfSync or Unknown states
+        description: >-
+          There are no apps in Synced state but there are one or more apps in OutOfSync or Unknown states
+        runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/gitops/deploy-from-git-to-k8s.md
--- a/rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
@@ -11,11 +11,9 @@ spec:
     rules:
     - alert: GitOpsApplicationSyncErrors
       expr: |
-        sum(last_over_time(argocd_app_info{sync_status="Synced"}[5m]))
+        sum(argocd_app_info{sync_status="Synced", namespace="gitops-service-argocd"})
         /
-        (
-          sum(last_over_time(argocd_app_info{sync_status="Synced"}[5m])) + sum(last_over_time(argocd_app_info{sync_status="OutOfSync"}[5m])) + sum(last_over_time(argocd_app_info{sync_status="Unknown"}[5m]))
-        ) < 0.95
+        sum(argocd_app_info{namespace="gitops-service-argocd"}) < 0.95
       for: 5m
       labels:
         severity: warning
--- a/test/promql/tests/data_plane/gitops_sync_test.yaml
+++ b/test/promql/tests/data_plane/gitops_sync_test.yaml
@@ -11,11 +11,11 @@ tests:
     input_series:
 
       - series: 'argocd_app_info{sync_status="Synced",name="gitopsdepl-001",namespace="gitops-service-argocd"}'
-        values: '_ _ _ _ _ _'
+        values: '_ _ _ _ _ _ _ _ _ _ _'
       - series: 'argocd_app_info{sync_status="OutOfSync",name="gitopsdepl-002",namespace="gitops-service-argocd"}'
-        values: '1 1 1 1 1 1'
+        values: '1 1 1 1 1 1 1 1 1 1 1'
       - series: 'argocd_app_info{sync_status="Unknown",name="gitopsdepl-003",namespace="gitops-service-argocd"}'
-        values: '1 1 1 1 1 1'
+        values: '1 1 1 1 1 1 1 1 1 1 1'
 
     alert_rule_test:
       - eval_time: 10m
@@ -36,9 +36,9 @@ tests:
 
       # We don't want to be alerted in this situation (don't care situation)
       - series: 'argocd_app_info{sync_status="Synced",name="gitopsdepl-001",namespace="gitops-service-argocd"}'
-        values: '_ _ _ _ _ _'
+        values: '_ _ _ _ _ _ _ _ _ _ _'
       - series: 'argocd_app_info{sync_status="OutOfSync",name="gitopsdepl-002",namespace="gitops-service-argocd"}'
-        values: '_ _ _ _ _ _'
+        values: '_ _ _ _ _ _ _ _ _ _ _'
 
     alert_rule_test:
       - eval_time: 6m
@@ -57,11 +57,11 @@ tests:
 
       # One synced app, but one of each of OutOfSync and Unknown. 33% sync rate, so alert
       - series: 'argocd_app_info{sync_status="Synced",name="gitopsdepl-001",namespace="gitops-service-argocd"}'
-        values: '1 1 1 1 1 1'
+        values: '1 1 1 1 1 1 1 1 1 1 1 1'
       - series: 'argocd_app_info{sync_status="OutOfSync",name="gitopsdepl-002",namespace="gitops-service-argocd"}'
-        values: '1 1 1 1 1 1'
+        values: '1 1 1 1 1 1 1 1 1 1 1 1'
       - series: 'argocd_app_info{sync_status="Unknown",name="gitopsdepl-003",namespace="gitops-service-argocd"}'
-        values: '1 1 1 1 1 1'
+        values: '1 1 1 1 1 1 1 1 1 1 1 1'
 
     alert_rule_test:
       - eval_time: 6m
@@ -82,11 +82,11 @@ tests:
 
       # Less than 95% synced rate (at 50%), should alert
       - series: 'argocd_app_info{sync_status="Synced",name="gitopsdepl-001",namespace="gitops-service-argocd"}'
-        values: '1 1 1 1 1 1'
+        values: '1 1 1 1 1 1 1 1 1 1 1'
       - series: 'argocd_app_info{sync_status="OutOfSync",name="gitopsdepl-002",namespace="gitops-service-argocd"}'
-        values: '1 1 1 1 1 1'
+        values: '1 1 1 1 1 1 1 1 1 1 1'
       - series: 'argocd_app_info{sync_status="Unknown",name="gitopsdepl-003",namespace="gitops-service-argocd"}'
-        values: '_ _ _ _ _ _'
+        values: '_ _ _ _ _ _ _ _ _ _ _'
 
     alert_rule_test:
       - eval_time: 6m
@@ -107,11 +107,11 @@ tests:
 
       # One synced app for entire window period. Basic scenario. Should not alert
       - series: 'argocd_app_info{sync_status="Synced",name="gitopsdepl-001",namespace="gitops-service-argocd"}'
-        values: '1 1 1 1 1 1'
+        values: '1 1 1 1 1 1 1 1 1 1 1'
       - series: 'argocd_app_info{sync_status="OutOfSync",name="gitopsdepl-002",namespace="gitops-service-argocd"}'
-        values: '_ _ _ _ _ _'
+        values: '_ _ _ _ _ _ _ _ _ _ _'
       - series: 'argocd_app_info{sync_status="Unknown",name="gitopsdepl-003",namespace="gitops-service-argocd"}'
-        values: '_ _ _ _ _ _'
+        values: '_ _ _ _ _ _ _ _ _ _ _'
 
     alert_rule_test:
       - eval_time: 6m
@@ -122,11 +122,11 @@ tests:
 
       # No apps. No alerts
       - series: 'argocd_app_info{sync_status="Synced",name="gitopsdepl-001",namespace="gitops-service-argocd"}'
-        values: '_ _ _ _ _ _'
+        values: '_ _ _ _ _ _ _ _ _ _ _'
       - series: 'argocd_app_info{sync_status="OutOfSync",name="gitopsdepl-002",namespace="gitops-service-argocd"}'
-        values: '_ _ _ _ _ _'
+        values: '_ _ _ _ _ _ _ _ _ _ _'
       - series: 'argocd_app_info{sync_status="Unknown",name="gitopsdepl-003",namespace="gitops-service-argocd"}'
-        values: '_ _ _ _ _ _'
+        values: '_ _ _ _ _ _ _ _ _ _ _'
 
     alert_rule_test:
       - eval_time: 6m
--- a/test/promql/tests/data_plane/gitops_sync_test.yaml
+++ b/test/promql/tests/data_plane/gitops_sync_test.yaml
@@ -10,11 +10,11 @@ tests:
   - interval: 1m
     input_series:
 
-      - series: 'argocd_app_info{sync_status="Synced",name="gitopsdepl-001",namespace="gitops-service-argocd",source_cluster="stonesoupp01ue1"}'
+      - series: 'argocd_app_info{sync_status="Synced",name="gitopsdepl-001",namespace="gitops-service-argocd"}'
         values: '_ _ _ _ _ _'
-      - series: 'argocd_app_info{sync_status="OutOfSync",name="gitopsdepl-002",namespace="gitops-service-argocd",source_cluster="stonesoupp01ue1"}'
+      - series: 'argocd_app_info{sync_status="OutOfSync",name="gitopsdepl-002",namespace="gitops-service-argocd"}'
         values: '1 1 1 1 1 1'
-      - series: 'argocd_app_info{sync_status="Unknown",name="gitopsdepl-003",namespace="gitops-service-argocd",source_cluster="stonesoupp01ue1"}'
+      - series: 'argocd_app_info{sync_status="Unknown",name="gitopsdepl-003",namespace="gitops-service-argocd"}'
         values: '1 1 1 1 1 1'
 
     alert_rule_test:
@@ -35,9 +35,9 @@ tests:
     input_series:
 
       # We don't want to be alerted in this situation (don't care situation)
-      - series: 'argocd_app_info{sync_status="Synced",name="gitopsdepl-001",namespace="gitops-service-argocd",source_cluster="stonesoupp01ue1"}'
+      - series: 'argocd_app_info{sync_status="Synced",name="gitopsdepl-001",namespace="gitops-service-argocd"}'
         values: '_ _ _ _ _ _'
-      - series: 'argocd_app_info{sync_status="OutOfSync",name="gitopsdepl-002",namespace="gitops-service-argocd",source_cluster="stonesoupp01ue1"}'
+      - series: 'argocd_app_info{sync_status="OutOfSync",name="gitopsdepl-002",namespace="gitops-service-argocd"}'
         values: '_ _ _ _ _ _'
 
     alert_rule_test:
@@ -56,11 +56,11 @@ tests:
     input_series:
 
       # One synced app, but one of each of OutOfSync and Unknown. 33% sync rate, so alert
-      - series: 'argocd_app_info{sync_status="Synced",name="gitopsdepl-001",namespace="gitops-service-argocd",source_cluster="stonesoupp01ue1"}'
+      - series: 'argocd_app_info{sync_status="Synced",name="gitopsdepl-001",namespace="gitops-service-argocd"}'
         values: '1 1 1 1 1 1'
-      - series: 'argocd_app_info{sync_status="OutOfSync",name="gitopsdepl-002",namespace="gitops-service-argocd",source_cluster="stonesoupp01ue1"}'
+      - series: 'argocd_app_info{sync_status="OutOfSync",name="gitopsdepl-002",namespace="gitops-service-argocd"}'
         values: '1 1 1 1 1 1'
-      - series: 'argocd_app_info{sync_status="Unknown",name="gitopsdepl-003",namespace="gitops-service-argocd",source_cluster="stonesoupp01ue1"}'
+      - series: 'argocd_app_info{sync_status="Unknown",name="gitopsdepl-003",namespace="gitops-service-argocd"}'
         values: '1 1 1 1 1 1'
 
     alert_rule_test:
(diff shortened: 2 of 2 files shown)


rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
```diff
@@ -0,0 +1,26 @@
...
+          expr: |
+            sum(argocd_app_info{health_status!="",sync_status=~"Synced"})
+            /
+            sum(argocd_app_info{health_status!="",sync_status!=""}) < 0.95
+          for: 5m
+          labels:
+            severity: warning
+          annotations:
+            summary: >-
+              Less than 95% of GitOps applications are not in synced state
```
yftacherzog, line 22: I think the `not` should not be there.
```suggestion
              Less than 95% of GitOps applications are in synced state
```

test/promql/tests/data_plane/gitops_sync_test.yaml
```diff
@@ -0,0 +1,84 @@
...
+
+rule_files:
+  - prometheus.gitops_sync_alerts.yaml
+
+tests:
+  # Alerted cases:
+  - interval: 1m
+    input_series:
+
+      # GitOps service is experiencing application sync failure rate of less than 95%, so alert
```
yftacherzog, line 11: Should say less than 95% success. No?
```suggestion
      # GitOps service is experiencing application sync success rate of less than 95%, so alert
```

test/promql/tests/data_plane/gitops_sync_test.yaml
```diff
@@ -0,0 +1,84 @@
...
+                The percentage total of all the Argo CD applications on cluster cluster01 having a sync
+                status of Synced is less than 95%.
+              runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/gitops/deploy-from-git-to-k8s.md
+
+  - interval: 1m
+    input_series:
+
+      # GitOps Service experienced less than 95% synced rate in applications, should alert
+      - series: 'argocd_app_info{sync_status="Synced", namespace="gitops-service-argocd", pod="pod-1", source_cluster="cluster01"}'
+        values: '1+1x60'
```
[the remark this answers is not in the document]
yftacherzog: https://prometheus.io/docs/prometheus/latest/configuration/unit_testing_rules/

test/promql/tests/data_plane/gitops_sync_test.yaml
```diff
@@ -0,0 +1,84 @@
...
+      - series: 'argocd_app_info{sync_status!="", namespace="gitops-service-argocd", pod="pod-1", source_cluster="cluster01"}'
+        values: '1+1x50 20+1x10'
+
+    alert_rule_test:
+      - eval_time: 65m
```
yftacherzog, line 44: We have 62m-worth of samples in this test (1 + 50 + 1 + 10), but we take the measurement at 65m. Unless we want to test the condition in which we don't have samples, we should have the evaluation happen within a time frame for which we have samples.

Also, I think we can do with 6m-worth of samples instead of ~1h. It will make things easier to comprehend.

test/promql/tests/data_plane/gitops_sync_test.yaml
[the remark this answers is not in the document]
keithchong: Hi @yftacherzog , how about the sum?  Wouldn't it be the sum of all the apps present?  In the case I was thinking of, an app is created for each interval.  So it will be like this:

1 2 3 4 5.....61

But then, at the 51st interval, 20 new apps are added but because of the above, only 1 of the apps will be in Synced state.

```
Number In Synced:   1 2 3 4 5.....49 50 51 52 53.....61
           Total:   1 2 3 4 5.....49 50 70 71 72.....82
```

At that point, the 95% criteria is not met, so there should be an alert.

Let me know if my interpretation is way off :-)

test/promql/tests/data_plane/gitops_sync_test.yaml
keithchong: Ah, it should have been 60, like the first test. It was a copy and paste from before the change.  I can reduce it down to 6m

keithchong: I committed a change for this and used the expanded notation instead. Also, the sequences are based on the sum and my interpretation of it, which from the other comment, might need to be changed.

test/promql/tests/data_plane/gitops_sync_test.yaml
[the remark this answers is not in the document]
yftacherzog: ok. I missed the part that the series created for the test was the sum rather than the raw data. It should be the raw data.

I also have some concerns about the validity of this query for this purpose. This is something that needs to be verified on a real environment to make sure that it behaves as expected.

My concern is like this: I think that the raw metric only gives value of 1 when the condition on the labels are satisfied and gives nothing when the conditions are not satisfied.

For example,  `argocd_app_info{sync_status="Synced", namespace="gitops-service-argocd", pod="pod-1", source_cluster="cluster01")}` will give 1 when it's in sync, but will not give 0 when it's out of sync.

I **suspect** that `argocd_app_info{health_status!="",sync_status=~"Synced"}` will keep returning the last value of the time series above, which is always 1.

So until the time series is cleaned out of the system (few days), this will keep returning 1.

This needs to be verified by forcing an app to be out of sync and checking if the sum changes.

I think the way around it will be to change the query to only take into account a time frame. e.g. `argocd_app_info{health_status!="",sync_status=~"Synced"}[5m]`, but that returns range vectors that you'd have to aggregate somehow, maybe with `last_over_time`.
e.g. `sum(last_over_time(argocd_app_info{health_status!="",sync_status=~"Synced"}[5m]))

All in all, the behavior needs to be verified on a proper environment and the unit test to be created in the end should be one that demonstrate how the data is present in Prometheus. I'd expect something like the following, but it needs to be verified:

```
      - series: 'argocd_app_info{sync_status="Synced", some_label="v1"}'
        values: '1 1 _ _ _ _ _ _'
      - series: 'argocd_app_info{sync_status="out-of-sync", some_label="v1"}'
        values: '_ _ 1 1 1 1 1 1'
      - series: 'argocd_app_info{sync_status="Synced", some_label="v2"}'
        values: '1 1 _ _ _ _ 1 1'
      - series: 'argocd_app_info{sync_status="out-of-sync", some_label="v2"}'
        values: '_ _ 1 1 1 1 _ _'
      - series: 'argocd_app_info{sync_status="Synced", some_label="v3"}'
        values: '1 1 1 1 1 1 1 1'
      - series: 'argocd_app_info{sync_status="out-of-sync", some_label="v3"}'
        values: '_ _ _ _ _ _ _ _'
```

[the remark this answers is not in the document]
keithchong: I've updated the tests and they run successfully

rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
```diff
@@ -0,0 +1,26 @@
...
+            description: >-
+              Less than 95% of GitOps applications in the namespace {{ $labels.namespace }} on cluster 
+              {{ $labels.source_cluster }} are not in Synced state.
```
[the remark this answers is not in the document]
keithchong: There appears to be an issue with using sum.  The pint linter complains that the query removes the namespace label even though the template has it.


```
INFO[0013] [rhobs/alerting/data_plane/prometheus.application_alerts.yaml] done 
FATA[0013] [rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml] failed to run 'pint'; output:
/tmp/obsctl-reloader-rules-checker3998111497/prometheus.gitops_sync_alerts.yaml:6-22 Bug: template is using "namespace" label but the query removes it (alerts/template)
  6 |           expr: |
  7 |             sum(argocd_app_info{sync_status="Synced"})
  8 |             /
  9 |             (
 10 |               sum(argocd_app_info{sync_status="Synced"}) +
 11 |               sum(argocd_app_info{sync_status="OutOfSync"}) +
 12 |               sum(argocd_app_info{sync_status="Unknown"})
 13 |             ) < 0.95
 ..
 21 |             description: >-
 22 |                 The percentage total of all Argo CD applications for namespace {{ $labels.namespace }} that are in Synced state is less than 95%.

level=info msg="Problems found" Bug=1
level=fatal msg="Execution completed with error(s)" error="found 1 problem(s) with severity Bug or higher"
 
make: *** [check-and-test] Error 1

```

rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
```diff
@@ -0,0 +1,28 @@
...
+  groups:
+  - name: gitops_sync_alerts
+    interval: 1m
+    rules:
+    - alert: GitOpsApplicationSyncErrors
+      expr: |
+        sum(last_over_time(argocd_app_info{sync_status="Synced"}[5m]))
+        /
+        (
+          sum(last_over_time(argocd_app_info{sync_status="Synced"}[5m])) + sum(last_over_time(argocd_app_info{sync_status="OutOfSync"}[5m])) + sum(last_over_time(argocd_app_info{sync_status="Unknown"}[5m]))
```
yftacherzog, line 17:
```suggestion
          sum(last_over_time(argocd_app_info{sync_status="Synced"}[5m])) + 
          sum(last_over_time(argocd_app_info{sync_status="OutOfSync"}[5m])) + 
          sum(last_over_time(argocd_app_info{sync_status="Unknown"}[5m]))
```

yftacherzog: I think this one doesn't necessarily sums up to the number of apps on the system. If an app changed state within the time frame, it will have last_over_time of 1 within the time frame for multiple statuses (as I don't think this metric ever returns something other than 1 - it either gives one or does not exist).

test/promql/tests/data_plane/gitops_sync_test.yaml
```diff
@@ -0,0 +1,86 @@
...
+  - prometheus.gitops_sync_alerts.yaml
+
+tests:
+  # Alerted cases:
+  - interval: 1m
+    input_series:
+
+      # GitOps service is experiencing application sync success rate of less than 95%, so alert
+      - series: 'argocd_app_info{sync_status="Synced"}'
+        values: '1 2 3 4 5 6'
```
yftacherzog, line 13: This metric cannot return anything but `1`. The possible values to have in test cases for this is either `1` or `_` (meaning, it was not scraped).

test/promql/tests/data_plane/gitops_sync_test.yaml
```diff
@@ -0,0 +1,86 @@
+evaluation_interval: 1m
+
+rule_files:
+  - prometheus.gitops_sync_alerts.yaml
+
+tests:
```
yftacherzog, line 6: In reality, this metric also has labels for identifying the apps for which the info is provided (namespace, name, etc.). We need test cases with some of those labels to see that the computations work as expected.

rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
```diff
@@ -0,0 +1,28 @@
...
+  name: rhtap-gitops-sync-alerting
+  labels:
+    tenant: rhtap
+spec:
+  groups:
+  - name: gitops_sync_alerts
+    interval: 1m
+    rules:
+    - alert: GitOpsApplicationSyncErrors
+      expr: |
```
yftacherzog, line 13: There's a corner case here to be tested and considered. What happens when no app is in `sync_status="Synced"`. I'm not sure the LHS of the operation will return 0 - it might not return anything. We need to have it tested and in case we have a problem there, we need to decide if we want to address it.

yftacherzog (changes requested):
I think there are issues here with both the rule and the tests (see the comments).

For the rule, unfortunately, the `argocd_app_info` metric is not very trivial to have "smart" rules for. We were trying to have something similar in the past, but it did not work out the way we planned.

Any of those computations for percentage or comparing between quantities of different groups of time series is very fragile because of the nature of this metric.

I think we need to dumb it down.

What about having an alert that will check whether an individual app is with `sync_status!="Synced"` for more than x minutes? This should be easy to implement and will let us know whenever something is down for a long time (dangling apps might be something to consider in that case).

keithchong:
> I think there are issues here with both the rule and the tests (see the comments).
> 
> For the rule, unfortunately, the `argocd_app_info` metric is not very trivial to have "smart" rules for. We were trying to have something similar in the past, but it did not work out the way we planned.
> 
> Any of those computations for percentage or comparing between quantities of different groups of time series is very fragile because of the nature of this metric.
> 
> I think we need to dumb it down.
> 
> What about having an alert that will check whether an individual app is with `sync_status!="Synced"` for more than x minutes? This should be easy to implement and will let us know whenever something is down for a long time (dangling apps might be something to consider in that case).

Could you elaborate on why it is not trivial and why it is fragile?

The visualization that we have uses the same metric, where we do a sum, and it looks correct.  For example, this panel [here](https://grafana-access-appstudio-grafana.apps.stone-stg-rh01.l2vh.p1.openshiftapps.com/d/85a02a9f44f0cb886730d7251cfdbe16/gitops-argocd?orgId=1&editPanel=106).  It shows the number of apps in the various states.

Seems like the alert rule can still have the same expression, but the test, in particular, the series, has limitations or restrictions.  eg. series is only in the form:

`<metric name>{<label name>=<label value>, ...}
`
so we can't have sufficient test coverage to cover the alert rule.

rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
keithchong: I added the test, and it passed.

test/promql/tests/data_plane/gitops_sync_test.yaml
keithchong: I added the namespace="gitops-service-argocd" to each expression

test/promql/tests/data_plane/gitops_sync_test.yaml
keithchong: I updated the tests to just use 1 and _

rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
keithchong: I reverted it back to the original expression that I had, for now.  I replied to your comment.

yftacherzog:
> Could you elaborate on why it is not trivial and why it is fragile?

This metric only has one value - 1, which is given when the app is at the specific state denoted by the combination of labels. This exact combination of labels will get no data in any other state.

It makes it fragile as it requires you to use queries that examine a specific time frame (e.g. [5m]). Otherwise, you don't know how old is the value you use, and at the same time, that time frame can have multiple mutually exclusive label combinations getting value 1 (not at the exact same time but at the same time window being considered).

As your rule was summarizing all those combinations together, it could actually reach more than 100%, making the computation wrong. I see that you changed it. I'll review again.

yftacherzog:
> so we can't have sufficient test coverage to cover the alert rule.

What sort of coverage are you missing?

test/promql/tests/data_plane/gitops_sync_test.yaml
yftacherzog: but they are all the same namespace. In reality, you have time series for multiple apps. Your current tests don't test the way you handle multiple apps.

test/promql/tests/data_plane/gitops_sync_test.yaml
```diff
@@ -0,0 +1,101 @@
...
+  - interval: 1m
+    input_series:
+
+      # GitOps service is experiencing application sync success rate of less than 95%, so alert
+      - series: 'argocd_app_info{sync_status="Synced",namespace="gitops-service-argocd"}'
+        values: '1 1 1 1 1 1'
+      - series: 'argocd_app_info{sync_status="OutOfSync",namespace="gitops-service-argocd"}'
+        values: '1 1 1 1 1 1'
+      - series: 'argocd_app_info{sync_status="Unknown",namespace="gitops-service-argocd"}'
+        values: '1 1 1 1 1 1'
```
yftacherzog, line 17: Those statuses should be mutually. In this case, the same app is in 3 different sync statuses at the same time.

test/promql/tests/data_plane/gitops_sync_test.yaml
yftacherzog: Moreover, the SLO itself, IIUC, is about summarizing multiple apps, so I can't see it serves purpose without providing additional time series for additional apps.

test/promql/tests/data_plane/gitops_sync_test.yaml
```diff
@@ -0,0 +1,101 @@
...
+              runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/gitops/deploy-from-git-to-k8s.md
+
+  - interval: 1m
+    input_series:
+
+      # GitOps Service experienced less than 95% synced rate in applications, should alert
+      - series: 'argocd_app_info{sync_status="Synced",namespace="gitops-service-argocd"}'
+        values: '1 1 1 1 1 1'
+      - series: 'argocd_app_info{sync_status="OutOfSync",namespace="gitops-service-argocd"}'
+        values: '1 1 1 1 1 1'
```
yftacherzog, line 40: Those are mutually exclusive.

test/promql/tests/data_plane/gitops_sync_test.yaml
```diff
@@ -0,0 +1,101 @@
...
+  - interval: 1m
+    input_series:
+
+      # GitOps Service experienced greater than 95% synced rate for applications, should not alert
+      - series: 'argocd_app_info{sync_status="Synced",namespace="gitops-service-argocd"}'
+        values: '1 1 1 1 1 1'
+      - series: 'argocd_app_info{sync_status="OutOfSync",namespace="gitops-service-argocd"}'
+        values: '_ _ 1 _ 1 _'
+      - series: 'argocd_app_info{sync_status="Unknown",namespace="gitops-service-argocd"}'
+        values: '_ _ _ 1 _ 1'
```
yftacherzog, line 82: I think (no sure), that if you'd make those time series a little longer, then you'd get an alert.

The reason being is that, for Prometheus, the value of the 2 non-synced time series is still `1` as due to the nature of this metric, it cannot get any other value other than `1`.

The way to overcome it is to have the alert rule examine specific time windows, but there we have issues as well.
```
      - series: 'argocd_app_info{sync_status="Synced",namespace="gitops-service-argocd"}'
      - series: 'argocd_app_info{sync_status="OutOfSync",namespace="gitops-service-argocd"}'
        values: '_ _ 1 _ 1 _ _ _ _ _ _'
      - series: 'argocd_app_info{sync_status="Unknown",namespace="gitops-service-argocd"}'
        values: '_ _ _ 1 _ 1 _ _ _ _ _'
```

There might be a way to do it, but there are many corner cases to consider here. 

One option that might be less fragile (I think it's an overkill) is to have a **recording rule** that will aggregate the combination of such time series, for each app, to a single metric that will provide 1 if it's synced and 0 otherwise. Then we can have an **alerting rule** utilizing that metric.

I'm not optimistic about the ability to make those computations based on the `argocd_app_info` metric (I'll be glad to be proven wrong).

avi-biton (commented):
If we will take for example the possibility that we have no synced_status="Synced"(assuming we got only failures)
In this case the rule defined here:
```
sum(argocd_app_info{sync_status="Synced", namespace="gitops-service-argocd"})
 /
sum(argocd_app_info{namespace="gitops-service-argocd"}) < 0.95
```
will produce an error.
the reason is that we do not have any time series that matches the query `sum(argocd_app_info{sync_status="Synced", namespace="gitops-service-argocd"})`  and we will get the error: 
`No datapoints found.`

test/promql/tests/data_plane/gitops_sync_test.yaml
keithchong: @yftacherzog , all your comments are enlightening.

> I think (not sure), that if you'd make those time series a little longer, then you'd get an alert.

It did not. I added additional tests to show how to get an alert.

> The reason being is that, for Prometheus, the value of the 2 non-synced time series is still 1 as due to the nature of this metric, it cannot get any other value other than 1.

In the tests, I made them distinct by name, so each series should have a value 1 or not. This seems to be confirmed by looking at the OpenShift console Metrics page.

I checked in another commit and included descriptive comments for each test.

test/promql/tests/data_plane/gitops_sync_test.yaml
keithchong: I added the distinct names

test/promql/tests/data_plane/gitops_sync_test.yaml
keithchong: I added distinct names

test/promql/tests/data_plane/gitops_sync_test.yaml
keithchong: From the GitOps service POV, I think all our apps are deployed to the same namespace. I can double check with Jonathan.

keithchong: I added multiple apps/series.  We can add more if needed.

keithchong:
@avi-biton I had to add a new rule and test GitOpsApplicationNoSyncApps for the situation where there are no Synced apps, but at least one or more apps that are in OutOfSync or Unknown states.

rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
```diff
@@ -0,0 +1,44 @@
...
+        severity: warning
+        slo: true
+      annotations:
+        summary: >-
+          Less than 95% of GitOps applications are in synced state
+        description: >-
+          The percentage total of all Argo CD applications that are in Synced state is less than 95%.
+        runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/gitops/deploy-from-git-to-k8s.md
+    - alert: GitOpsApplicationNoSyncApps
+      expr: |
```
yftacherzog, line 28: The use of the `absent_over_time` (or maybe `absent`) function might be more straightforward.
https://prometheus.io/docs/prometheus/latest/querying/functions/#absent_over_time

test/promql/tests/data_plane/gitops_sync_test.yaml
```diff
@@ -0,0 +1,249 @@
...
+              summary: >-
+                Less than 95% of GitOps applications are in synced state
+              description: >-
+                The percentage total of all Argo CD applications that are in Synced state is less than 95%.
+              runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/gitops/deploy-from-git-to-k8s.md
+
+  - interval: 1m
+    input_series:
+
+      # One synced app for entire window perios. Basic scenario. Should not alert
```
yftacherzog, line 88: typo `perios`

test/promql/tests/data_plane/gitops_sync_test.yaml
yftacherzog: It behaves a little different than what I thought, but still there's a delicate point here.

Contrary to what I suggested above, the latest timestamp being considered needs to be within the window, but Prometheus will assume it to be unchanged until it gets a new value (which it never will for this metric). It means that the 5m window we give is only accumulating noise rather than eliminating it.

The test below does raise the alert (I don't think that's the rule's desired behavior).
It has a single app that was out of sync for 1 minute and in sync for the rest of the time window. The `1` value provided when it was out of sync is still being considered as long as it's still within the rule's window (5m), and while we should be 100% in sync, we're actually only 50% in sync and we get an alert. 

```
      - series: 'argocd_app_info{sync_status="Synced",name="gitopsdepl-001",namespace="gitops-service-argocd"}'
        values: ' 1 _ 1 1 1 1'
      - series: 'argocd_app_info{sync_status="OutOfSync",name="gitopsdepl-001",namespace="gitops-service-argocd"}'
        values: ' _ 1 _ _ _ _'

    alert_rule_test:
      - eval_time: 6m
        alertname: GitOpsApplicationSyncErrors
```

rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
avi-biton: The test you need to add is for `GitOpsApplicationSyncErrors` with no apps in `sync_status="Synced`
This test would probably fail because:
`sum(argocd_app_info{sync_status="Synced", namespace="gitops-service-argocd"})` with values like: 
```
- series: 'argocd_app_info{sync_status="Synced",namespace="gitops-service-argocd"}'
        values: '1 1 1 1 1 1'
```
Will return `No datapoints found` and we will not be able to divide it with the next expression:
`sum(argocd_app_info{namespace="gitops-service-argocd"})`

rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
```diff
@@ -0,0 +1,44 @@
...
+        (
+          sum(argocd_app_info{sync_status="OutOfSync", namespace="gitops-service-argocd"}) +
```
avi-biton, line 30: We will get no answer if we have no applications with `sync_status="OutOfSync"` or `sync_status="Unknown"`
We will get  `No datapoints found`

rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
```diff
@@ -0,0 +1,44 @@
...
+  labels:
+    tenant: rhtap
+spec:
+  groups:
+  - name: gitops_sync_alerts
+    interval: 1m
+    rules:
+    - alert: GitOpsApplicationSyncErrors
+      expr: |
+        sum(argocd_app_info{sync_status="Synced", namespace="gitops-service-argocd"})
```
yftacherzog, line 14: It might be that you can overcome the corner case of no synced apps by replacing this with something like this (I didn't try):
```suggestion
        (
        sum(argocd_app_info{sync_status="Synced", namespace="gitops-service-argocd"})
        or sum(argocd_app_info{namespace="gitops-service-argocd"}) * 0
        )
```

This should replace the missing sum with a time series with the same labels a value of zero.

See a reference here: https://promlabs.com/blog/2023/09/13/dealing-with-missing-time-series-in-prometheus/
and a reference to the `or` operator: https://prometheus.io/docs/prometheus/latest/querying/operators/#logical-set-binary-operators

test/promql/tests/data_plane/gitops_sync_test.yaml
yftacherzog: Maybe (really maybe) we can mitigate that with evaluating range vectors of small range (say 1m) over the 5 minutes period. This should allow us to have the alert wait for few minutes before firing while considering short time windows so the issue we see above is somewhat diminished a little.

Something like this (needs to be combined with my suggestion for eliminating the no-synced scenario:
```
      expr: |
        sum(
          last_over_time(
            argocd_app_info{sync_status="Synced", namespace="gitops-service-argocd"}[1m]
          )
        )
        /
        sum(
          last_over_time(
            argocd_app_info{namespace="gitops-service-argocd"}[1m]
          )
        )
        < 0.95
      for: 5m
```

rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
```diff
@@ -0,0 +1,44 @@
...
+        sum(argocd_app_info{namespace="gitops-service-argocd"}) < 0.95
+      for: 5m
+      labels:
+        severity: warning
+        slo: true
+      annotations:
+        summary: >-
+          Less than 95% of GitOps applications are in synced state
+        description: >-
+          The percentage total of all Argo CD applications that are in Synced state is less than 95%.
```
yftacherzog, line 25: You can include the actual value of the computation within those fields. It should at least help you troubleshooting (might also be informative for the actual alerts).

Include `{{ $value }}` in the summary/description fields.

rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
```diff
@@ -0,0 +1,44 @@
...
+  name: rhtap-gitops-sync-alerting
+  labels:
+    tenant: rhtap
+spec:
+  groups:
+  - name: gitops_sync_alerts
+    interval: 1m
+    rules:
+    - alert: GitOpsApplicationSyncErrors
+      expr: |
```
yftacherzog, line 13: The current expression is summarizing the apps across all env clusters (e.g. all staging host and member clusters). Is this the expected behavior?

Suppose that we have multiple apps out of sync in one stage member cluster. Do we want the alert to be fired when they are more than 5% of the apps within that cluster or between all stage clusters?

yftacherzog: See the use of `without` and `by` here: https://prometheus.io/docs/prometheus/latest/querying/operators/#aggregation-operators

rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
avi-biton: In this case we might get a value of `0` which if fine as it is, but we are dividing this `0` with the result of another query.
We most likely get an error for this division.

yftacherzog, line 14: I don't see any time series with `namespace="gitops-service-argocd"` in AppSRE Grafana. Which apps are we expecting to be on that namespace?

rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
keithchong: Latest commit includes $value

test/promql/tests/data_plane/gitops_sync_test.yaml
keithchong: Done

rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
```diff
@@ -0,0 +1,52 @@
...
+        (
+          sum(avg_over_time(argocd_app_info{sync_status="Synced", namespace="gitops-service-argocd"}[5m]))
+          or
+          sum(argocd_app_info{namespace="gitops-service-argocd"}) * 0
+        )
+        /
+        (
+          sum(avg_over_time(argocd_app_info{namespace="gitops-service-argocd"}[5m]))
+          or
+          sum(argocd_app_info{namespace="gitops-service-argocd"}) * 0
```
yftacherzog, line 23: I suppose this expression is just adding clutter and only results with dividing by zero. I think we can remove it.

rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
yftacherzog: @keithchong is summing across all clusters the expected behavior?

rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
```diff
@@ -0,0 +1,52 @@
...
+        ) < 0.95
+      for: 5m
+      labels:
+        severity: warning
+        slo: true
+      annotations:
+        summary: >-
+          Less than 95% of GitOps applications are in synced state: {{ $value }}
```
yftacherzog, line 31: This can make it nicer if you prefer:
```suggestion
          Less than 95% of GitOps applications are in synced state: {{ $value | humanizePercentage }}
```

rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
```diff
@@ -0,0 +1,52 @@
...
+        description: >-
+          The percentage total of all Argo CD applications that are in Synced state is less than 95%.
+        runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/gitops/deploy-from-git-to-k8s.md
+    - alert: GitOpsApplicationNoSyncApps
```
yftacherzog, line 35: Do we still need this alert?
If I recall correctly, we only added it to overcome the state in which we have zero sync app. The current query for the other rule should have it covered now.

keithchong: Ah yes, the first four tests showed that we don't need it too.  The first two tests are the same as the third and fourth.  I'll remove this.

rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
keithchong: Yes, I prefer this  👍 , will make the change.

rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
keithchong: Removed.

rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
keithchong: Checked in a commit to address three of today's review comments

test/promql/tests/data_plane/gitops_sync_test.yaml
```diff
@@ -0,0 +1,666 @@
...
+  - prometheus.gitops_sync_alerts.yaml
+
+tests:
+  # -----------------------------------------------------------------------------------
+  # The following section describes tests for the situation where no apps are Synced
+  # -----------------------------------------------------------------------------------
+  - interval: 1m
+    input_series:
+
+      # Compare this test with the one for GitOpsApplicationNoSyncApps above
```
yftacherzog, line 13: This comment is referencing a non-existing alert

keithchong: Good catch. Comment removed.

rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
keithchong: @yftacherzog , I'm not sure if I can differentiate by cluster? Do you know what label to use?

yftacherzog: `source_cluster`.

Have a look at that metrics in https://grafana.app-sre.devshift.net/
It won't be your time series or clusters, but the label is the same.

test/promql/tests/data_plane/gitops_sync_test.yaml
```diff
@@ -0,0 +1,665 @@
...
+
+      - series: 'argocd_app_info{sync_status="Synced",name="gitopsdepl-001",namespace="gitops-service-argocd"}'
+        values: '_ _ _ _ _ _'
+      - series: 'argocd_app_info{sync_status="OutOfSync",name="gitopsdepl-002",namespace="gitops-service-argocd"}'
+        values: '1 1 1 1 1 1'
+      - series: 'argocd_app_info{sync_status="Unknown",name="gitopsdepl-003",namespace="gitops-service-argocd"}'
+        values: '1 1 1 1 1 1'
+
+    alert_rule_test:
+      - eval_time: 6m
```
yftacherzog, line 21: You have a look-back in the query of 5 minutes, and you're checking if it holds for 5 minutes. The eval time should be at least 10m.

yftacherzog:
@keithchong I think we need to better define what it means that an app is in sync.
Suppose that we have this test case:
```
- series: 'argocd_app_info{sync_status="Synced",name="gitopsdepl-001",namespace="gitops-service-argocd"}'
   values: '1 _ 1 _ 1 _ 1 _ 1 _'
- series: 'argocd_app_info{sync_status="OutOfSync",name="gitopsdepl-001",namespace="gitops-service-argocd"}'
   values: '_ 1 _ 1 _ 1 _ 1 _ 1'
```
We have a single app, but it will show as both in sync and out of sync and the result will be 50% availability, although as I understand it, it should be either 0% or 100%.

I think something is not entirely defined.

test/promql/tests/data_plane/gitops_sync_test.yaml
keithchong: @yftacherzog , could you please elaborate on this?  The interval is 1m

rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
keithchong: It looks like we might need a new rule for each cluster then, because I have to hardcode the name into the rule.

test/promql/tests/data_plane/gitops_sync_test.yaml
yftacherzog: The query is looking at the last 5 minutes and the alert is triggering after 5 minutes of the query being met. 

So the query returns valid results only after 5 minutes and then you need that to hold for additional 5 minutes for the alert to kick in.

rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
yftacherzog: You just need the label name rather than its value to include or exclude it from aggregation.

test/promql/tests/data_plane/gitops_sync_test.yaml
keithchong: I tried the above but it focusses on a  very narrow window and it depends on the when the evaluation occurs.  So it can alert when the app is out of sync for that specific window, even though the app can be in sync for several intervals.

I tried `avg_over_time` over `[5m]` and `for 5m`, which expands the window.

My test cases does show that it makes sense to do this, and probably explains why I'm getting 50%, when you're expecting to get 0%.

keithchong:
> @keithchong I think we need to better define what it means that an app is in sync. Suppose that we have this test case:
> 
> ```
> - series: 'argocd_app_info{sync_status="Synced",name="gitopsdepl-001",namespace="gitops-service-argocd"}'
>    values: '1 _ 1 _ 1 _ 1 _ 1 _'
> - series: 'argocd_app_info{sync_status="OutOfSync",name="gitopsdepl-001",namespace="gitops-service-argocd"}'
>    values: '_ 1 _ 1 _ 1 _ 1 _ 1'
> ```
> 
> We have a single app, but it will show as both in sync and out of sync and the result will be 50% availability, although as I understand it, it should be either 0% or 100%.
> 
> I think something is not entirely defined.

Please see my comment regarding the use of avg_over_time to explain why I'm getting 50%.

test/promql/tests/data_plane/gitops_sync_test.yaml
keithchong: Ok. I've widened the range, in the latest commit "Widen time series range". Resolving conversation.

test/promql/tests/data_plane/gitops_sync_test.yaml
yftacherzog: The only value this metric is generating is `1`, so average over time will be the same as max over time (or min over time), and for that reason, I think the use of average here is misleading.

The reason we're getting 50% is that the `argocd_app_info{sync_status="Synced"}` returns one series over that period and `argocd_app_info` returns two series over that period (the synced and the out of sync). Each of those have a value of `1`, so the sum of the synced is `1` and the sum of both synced and out of sync is `2`, and this is why we're getting `1/2`.

rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
yftacherzog: Need to include the source_cluster in some test cases as well to see it's doing its part.

test/promql/tests/data_plane/gitops_sync_test.yaml
yftacherzog: I'd say again what I said at the very beginning, I don't think the `argocd_app_info` metric is a good fit for measuring this SLO. My suggestion is to either change the SLO or to think of another way to measure it.

keithchong: I can switch it back to last_over_time over [1m].  I think this is the best we can do.  Perhaps we will not have the precise or perfect expression/formula.  The measurement is actually 'arbitrary' for a given window anyway, where the application sync state can dramatically change outside of that window.  The other perspective in favour of having the 1m window is a quick 'snapshot' of the system (regardless of how old a given series value is).  Trying out all the different test scenarios showed that for those situations where the sync state is stable (not changing over a reasonable period of time), we do get the expected results to alert or not to alert.  Typically, when an app is not synced or is unknown, it will likely stay that way until there is intervention.  eg. see the test cases where I had 20 apps and 10 apps.  We can also tweak the acceptance rate of 95% and lower it if we are getting alerted too often

rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
keithchong: I've added the test cases.  Note I had to update the alert rule as well otherwise, the first new test case that I added will fail.

rhobs/alerting/data_plane/prometheus.gitops_sync_alerts.yaml
keithchong: Naive question here, if we have a division by 0 in any situation, that means there are no apps at all.  So even if there is an error, isn't this a don't care situation, and we don't want alerts at all anyway?

yftacherzog (approved):

test/promql/tests/data_plane/gitops_sync_test.yaml
yftacherzog: ok

PR #97, opened 2023-10-30
chore(DEVHAS-481): Add alerting rules for the Application creation SLO

johnmcollier:
This PR adds alerting rules for the Application creation SLO as found [here](https://docs.google.com/document/d/1OVVZeCmv6o2qsZ1BhPaVPVzkJSa8cxLAN5ImF1eugOE/edit#bookmark=id.78n0do63qmep) (95% of created applications should have succeeded)

Corresponding alert tests have also been added

### DIFF
--- a/rhobs/alerting/data_plane/prometheus.application_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.application_alerts.yaml
@@ -11,10 +11,9 @@ spec:
     rules:
     - alert: ApplicationDeletionErrors
       expr: |
-        (sum(increase(has_application_failed_deletion_total[1h]))
-        by(namespace, pod, source_cluster) /
-        sum(increase(has_application_deletion_total[1h]))
-        by(namespace, pod, source_cluster)) > 0.01
+        (increase(has_application_failed_deletion_total[1h]))
+        /
+        (increase(has_application_deletion_total[1h]))  > 0.01
       for: 5m
       labels:
         severity: warning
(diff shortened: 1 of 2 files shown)


rhobs/alerting/data_plane/prometheus.application_alerts.yaml
```diff
@@ -27,3 +27,21 @@ spec:
           {{ $labels.namespace }} on cluster {{ $labels.source_cluster }} is failing to
           successfully delete at least 99% of applications over the past hour
         runbook_url: TBD
+    - alert: ApplicationCreationErrors
+      expr: |
+        (sum(increase(has_application_failed_creation_total[1h]))
+        by(namespace, pod, source_cluster) /
+        sum(increase(has_application_creation_total[1h]))
+        by(namespace, pod, source_cluster)) > 0.05
```
yftacherzog, line 35: Do the calls to `sum` here make any difference to the result of the individual queries? Can the combination of namespace/pod/source_cluster can potentially return more than a single time series?

yftacherzog: If not, then we should drop it. If it does, then we need test cases for that.

johnmcollier: @yftacherzog Thanks for the review. I've updated the alert accordingly. I also updated the other alert for Applications, as it used the same format.

yftacherzog (approved):
/lgtm

Kousalya1998 (approved):

PR #106, opened 2023-11-10
chore(PLNSRVCE-1427,PLNSRVCE-1428): Add PipelineRun SLO alert rules

gabemontero:


gabemontero:
WIP / do-not-merge-yet - The alert is not the final query I want to employ, but I wanted to start getting some validations from CI.

gabemontero:
```
Unit Testing:  /tmp/obsctl-reloader-rules-checker1288604589/pipeline_overhead_test.yaml\n
FAILED:\n
alertname: HighSchedulingOverhead, time: 13h, \n
exp:[\n
0:\n
Labels:{alertname=\"HighSchedulingOverhead\", severity=\"warning\", slo=\"true\"}\n
Annotations:{description=\"Tekton controller is requiring more than 5% of PipelineRun execution time to receive creation events.\", runbook_url=\"TBD\", summary=\"Tekton has a scheduling overhead vs. PipelineRun durations of >5%\"}\n
], \n
got:[]

\n\n
alertname: HighSchedulingOverhead, time: 13h, \n
exp:[\n
0:\n
Labels:{alertname=\"HighSchedulingOverhead\", severity=\"warning\", slo=\"true\"}\n
Annotations:{description=\"Tekton controller is requiring more than 5% of PipelineRun execution time to receive creation events.\", runbook_url=\"TBD\", summary=\"Tekton has a scheduling overhead vs. PipelineRun durations of >5%\"}\n
], \n
got:[]\n\n

alertname: HighExecutionOverhead, time: 13h, \n
exp:[\n

0:\n
Labels:{alertname=\"HighExecutionOverhead\", severity=\"warning\", slo=\"true\"}\n
Annotations:{description=\"Tekton controller is requiring more than 5% of PipelineRun execution time moving from one TaskRun to the next withing the PipelineRun\", runbook_url=\"TBD\", summary=\"Tekton has a execution overhead vs. PipelineRun durations of >5%\"}\n
], \n
got:[]\n\n


alertname: HighExecutionOverhead, time: 13h, \n

exp:[\n            0:\n

Labels:{alertname=\"HighExecutionOverhead\", severity=\"warning\", slo=\"true\"}\n
Annotations:{description=\"Tekton controller is requiring more than 5% of PipelineRun execution time moving from one TaskRun to the next withing the PipelineRun\", runbook_url=\"TBD\", summary=\"Tekton has a execution overhead vs. PipelineRun durations of >5%\"}\n
], \n

got:[]\n\n\n\n"


make: *** [Makefile:23: check-and-test] Error 1
```

gabemontero:
OK I've got a clean unit test run, but I had to comment out one of the 6 alert tests I added, and so far, am a bit stumped as to why it is failing.

gabemontero:
OK I've sorted it out wrt meets min understanding of https://prometheus.io/docs/prometheus/latest/configuration/unit_testing_rules/ and have clean run

removing WIP do-not-merge indicators, soliciting review for merge

gabemontero:
hey @amisstea can you either review or delegate as you deem appropriate for adding the pipeline svc alerts?  thanks

rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
```diff
@@ -0,0 +1,51 @@
...
+        - alert: HighSchedulingOverhead
+          expr: |
+            (sum(sum_over_time(pipelinerun_duration_scheduled_seconds_sum{status='succeded'}[30m]))
+            /
+            sum(sum_over_time(pipelinerun_duration_scheduled_seconds_count{status='succeded'}[30m])))
+            /
+            (sum(sum_over_time(tekton_pipelines_controller_pipelinerun_duration_seconds_sum{status='success'}[30m]))
+            /
+            sum(sum_over_time(tekton_pipelines_controller_pipelinerun_duration_seconds_count{status='success'}[30m])))  > 0.05
+          for: 12h
```
gabemontero, line 21: fyi this duration is intentiional ... we expect bumps in overhead from time to time, but want to see prolonged durations of higher overload before we start increasing tekton's minimal CPU/mem requirements or adding replicas of the controller/webhook

rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
```diff
@@ -0,0 +1,51 @@
...
+      rules:
+        - alert: HighSchedulingOverhead
+          expr: |
+            (sum(sum_over_time(pipelinerun_duration_scheduled_seconds_sum{status='succeded'}[30m]))
+            /
+            sum(sum_over_time(pipelinerun_duration_scheduled_seconds_count{status='succeded'}[30m])))
+            /
+            (sum(sum_over_time(tekton_pipelines_controller_pipelinerun_duration_seconds_sum{status='success'}[30m]))
+            /
+            sum(sum_over_time(tekton_pipelines_controller_pipelinerun_duration_seconds_count{status='success'}[30m])))  > 0.05
```
yftacherzog, line 20: I think indenting the brackets might make it more readable:
```suggestion
            (
                sum(sum_over_time(pipelinerun_duration_scheduled_seconds_sum{status='succeded'}[30m]))
                /
                sum(sum_over_time(pipelinerun_duration_scheduled_seconds_count{status='succeded'}[30m]))
            )
            /
            (
                sum(sum_over_time(tekton_pipelines_controller_pipelinerun_duration_seconds_sum{status='success'}[30m]))
                /
                sum(sum_over_time(tekton_pipelines_controller_pipelinerun_duration_seconds_count{status='success'}[30m]))
            )  > 0.05
```

rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
```diff
@@ -0,0 +1,51 @@
...
+          for: 12h
+          labels:
+            severity: warning
+            slo: true
+          annotations:
+            summary: >-
+              Tekton has a execution overhead vs. PipelineRun durations of >5%
```
yftacherzog, line 47:
```suggestion
              Tekton has an execution overhead vs. PipelineRun durations of >5%
```

yftacherzog: You can use `{{ $value }}` to also show the calculated percentage.

rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
```diff
@@ -0,0 +1,51 @@
...
+  name: rhtap-pipeline-alerting
+  labels:
+    tenant: rhtap
+spec:
+  groups:
+    - name: pipeline_alerts
+      interval: 1m
+      rules:
+        - alert: HighSchedulingOverhead
+          expr: |
```
yftacherzog, line 13: We're summarizing across clusters. Is this the intention?

yftacherzog: I think we should use either `increase` 
 (https://prometheus.io/docs/prometheus/latest/querying/functions/#increase) or maybe `histogram_count` and `histogram_sum` (https://prometheus.io/docs/prometheus/latest/querying/functions/#histogram_count-and-histogram_sum). I Never tried the latter.

If I understand correctly, we currently assume that the metrics are monotonically growing, but that might not always be the case.
The use of, at least, `increase` should help overcoming counter resets etc.

test/promql/tests/data_plane/pipeline_overhead_test.yaml
```diff
@@ -0,0 +1,159 @@
+evaluation_interval: 1m
+
+rule_files:
+  - prometheus.pipeline_alerts.yaml
+
+tests:
```
yftacherzog, line 6: We're using sum, but we're not testing any summing.

rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
yftacherzog: I think we should use either `increase` 
 (https://prometheus.io/docs/prometheus/latest/querying/functions/#increase) or maybe `histogram_count` and `histogram_sum` (https://prometheus.io/docs/prometheus/latest/querying/functions/#histogram_count-and-histogram_sum). I Never tried the latter.

If I understand correctly, we currently assume that the metrics are monotonically growing, but that might not always be the case.
The use of, at least, `increase` should help overcoming counter resets etc.

(to be honest, I can't say I fully understand the metrics yet, but I tend to think that `sum_over_time` might not be the way to go here, as `_sum` and `_count` should behave as counters -- as far as I understand)

gabemontero: Unintentional.

I do see `{{ $labels.source_cluster }}` in the descriptions of some of the existing alerts.  Is that how I make it "per cluster" @yftacherzog ?

rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
gabemontero: ok, and put that in the description, correct?   I'll add that in my next push to the branch/PR thanks

rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
gabemontero: On the second point, 
- simply using sum was not sufficient through manual testing with the OCP metric console with these histograms.  The combination was needed.
- that said, yes, my observation over many weeks has been that the metric only grows until the controller I'm using to measure is restarted.  If that is what you mean by a reset, at least with the 30 minute scrape window, that had been a non issue.
- I will research the increase / histogram_count / histogram_sum functions and if what they produces reconciles with me, I'll switch to those.  If they do not reconcile, I'll describe what is wrong, and we can continue from there.

test/promql/tests/data_plane/pipeline_overhead_test.yaml
gabemontero: I'm new to this prom unit testing, but from what I've learned so far, it is summing the 781 number of entries I have specified.  But I am probably missing some nuance which you are getting at @yftacherzog .... could you please elaborate by citing one of my input_series values, and why it is not summing?  thanks

rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
yftacherzog: > Unintentional.
> 
> I do see `{{ $labels.source_cluster }}` in the descriptions of some of the existing alerts. Is that how I make it "per cluster" @yftacherzog ?

`{{ $labels.source_cluster }}` is the label that tells us the where the metric came from. We should have it in the summary/description, but more importantly, the query we did should make the calculations per-cluster (assuming that is what we're after). It means that if we use `sum` then we need to make sure time series with different source_cluster are not summed together.

yftacherzog: > On the second point,
> 
> * simply using sum was not sufficient through manual testing with the OCP metric console with these histograms.  The combination was needed.
> * that said, yes, my observation over many weeks has been that the metric only grows until the controller I'm using to measure is restarted.  If that is what you mean by a reset, at least with the 30 minute scrape window, that had been a non issue.
> * I will research the increase / histogram_count / histogram_sum functions and if what they produces reconciles with me, I'll switch to those.  If they do not reconcile, I'll describe what is wrong, and we can continue from there.

Yes. I meant counter reset due to something like controller restart. There are certain query functions Prometheus offers (e.g. `increase`) that should be able to detect counter resets and give a correct value even when the counter resets.

test/promql/tests/data_plane/pipeline_overhead_test.yaml
yftacherzog: `sum` is a function that aggregates the **current value** across a vector of time series. It will not aggregate over time (this is `sum_over_time` which is aggregating over a range vector. e.g. my-metric[5m])

suppose that you have this:
foo{bar="a"}
1 1 1 1
foo{bar="b"}
1 1 1 1
foo{bar="c"}
1 1 1 1

`sum(foo)` will give us:
foo
3 3 3 3

So, suppose this is what we're after, then we need test cases in which we have more than one time series that fits the queries that we do to see that we're summing the way we expect.

rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
yftacherzog: indeed

rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
gabemontero: So I logged onto the RHTAP Prod RH cluster at https://console-openshift-console.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com and I went to Observer-> Metrics

1) neither `histogram_count` nor `histogram_sum` were available as functions; my first assumption there is that those clusters are not a a sufficient level wrt prometheus to container those features.  I'm not sure how the flow from the App SRE grafana to the individual clusters goes, and where those 2 functions are used, but it seem like if it does not work on the local cluster, those are not available to us.
2) `increase` does exist.... after analyzing what `increase(...)` produces I was able to modify one of our overhead queries where I replaced `sum_over_time` with increase, and get an scheduling overhead ratio that was believable

```
sum(increase(pipelinerun_duration_scheduled_seconds_sum{status='succeded'}[30m])) / sum(increase(pipelinerun_duration_scheduled_seconds_count{status='succeded'}[30m])) / sum(increase(tekton_pipelines_controller_pipelinerun_duration_seconds_sum{status='success'}[30m])) / sum(increase(tekton_pipelines_controller_pipelinerun_duration_seconds_count{status='success'}[30m])) * 100
```

So minimally, I'm fine with replacing `sum_over_time` with `increase`.  Now, of course, you see I still need to use `sum` in order to sufficiently aggregate the various _sum and _count observations of the histograms we are using in order to calculate an overall percentage.

However, wrt using `sum` and making sure different source clusters are not summed together, I can only make educated guesses as to how to filter.  Do I add a `cluster` label in the query?  Or is there something else?  Also, in looking at the other existing alerts, I see no such cluster filtering, and am unclear as to why the use of `sum` makes any difference wrt using data from multiple clusters?.  Aren't those other queries technically being run against multiple clusters?

Ultimately the alerts are defined on the individual clusters, no?  Or is this a question of the App SRE grafana dashbaord running outside our clusters, and being able to query against multilpe clusters, and hence we want what the app sre dashboard shows to match what the individual cluster is alerting on?

Lastly, eventually, RHTAP is suppose to go down to one production cluster, so it might ultimately be a moot point.  But ideally we sort out the above in case that ever changes.

Thanks

gabemontero: Or should I be using something other than `sum` to take all the entries in the vectors produced by `increase` into single values ?

yftacherzog: The alerts are managed in RHOBS which contains metrics coming from all data plane (and control plane clusters).

When you aggregate (e.g. with `sum`), you can specify according to which labels you're going to aggregate. you do that using the `by` and `without` keywords. It will make Prometheus keep only or everything-but the labels you specify. I think (always get confused with that) you'd need to use something like:
```
sum by (source_cluster) (my_metric{my_label="foo", another_label="bar"...})
```
in this case. Give it a try and see what you get back.

See this reference for more details: https://prometheus.io/docs/prometheus/latest/querying/operators/#aggregation-operators

About whether or not we should sum the clusters together, I leave it for you to decide.

test/promql/tests/data_plane/pipeline_overhead_test.yaml
yftacherzog: I think you should at least have more than one time series that should be **included** in the sum. one which should be excluded would be a nice addition.

rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
gabemontero: OK thanks for the pointer.  Back over at https://console-openshift-console.apps.stone-stg-m01.7ayg.p1.openshiftapps.com and the metrics browser with its auto-completion I discovered that `sum by (cluster_id)` is a thing and it works correctly when I add it to my queries as is (i.e. still `sum_over_time` vs. `increase`)

`make check-and-test ` also passed when I added it to the alerts.

Now, given the `source_cluster` label already noted above, I'm guessing that when done from RHOBS and not a single cluster's metrics browser, I should use `source_cluster` as you noted vs. `cluster_id`.  `make check-and-test` passed with that as well.

Regardless of what I did with `sum...` when I switched my unit tests from `sum_over_time` to `increase` the `make check-and-test` started to fail.  I'll see if I can sort that out as I manipulate the time series in the tests per the thread below.

gabemontero:
Hmmm ... something is up now with RHTAP RH Prod cluster or with quay.  None of the clone-repository, etc. tasks can be resolved

Screenshot at 2023-11-14 15-19-12


I'll pursue that separately.  Short term @yftacherzog I *believe* I've addressed various feedback
- using {{ value }} and {{ labels.source_cluster }} in the alert text
- filtering sums by source_cluster
- switched from sum_over_time to increase
- adjusted tests accordingly, and added additional series to better vet the use of sum

PTAL when you get the chance / htanks

gabemontero:
yep quay issues re https://status.quay.io/ as I type and https://redhat.enterprise.slack.com/archives/C02CTEB3MMF/p1699993906601559

yftacherzog:
@gabemontero Interestingly enough, when I now run the query for the first rule in AppSRE Grafana for the production data source, I get results for only one cluster. 

I think the reason for that is that metrics `tekton_pipelines_controller_pipelinerun_duration_seconds_sum` and `tekton_pipelines_controller_pipelinerun_duration_seconds_count` only have results from cluster `stone-prd-rh01`.

Two questions:

1. Any idea why? Is that ok?
2. Would that be the expected result for your query in that case or no entries is a scenario the rule should cover?

rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
```diff
@@ -0,0 +1,57 @@
...
+            )  > 0.05
+          for: 12h
+          labels:
+            severity: warning
+            slo: true
+          annotations:
+            summary: >-
+              Tekton has a scheduling overhead vs. PipelineRun durations of >5%
+            description: >-
+              Tekton controller on cluster {{ $labels.source_cluster }} has a ratio of the time needed to receive PipelineRun creation events vs. overall PipelineRun execution time of {{ $value }} instead of less than 0.05.
```
yftacherzog, line 33: Can we break this line for readability (same for the other rule)?

yftacherzog: For the percentage value, if you want, can also use `{{ $value | humanizePercentage }}` as @kim-tsao did here: https://github.com/redhat-appstudio/o11y/pull/108/files#diff-122adf1739691203cb3c64d8e965e4def93cab660ca866a3456ad9a9c787e3b6R23

gabemontero:
> @gabemontero Interestingly enough, when I now run the query for the first rule in AppSRE Grafana for the production data source, I get results for only one cluster.
> 
> I think the reason for that is that metrics `tekton_pipelines_controller_pipelinerun_duration_seconds_sum` and `tekton_pipelines_controller_pipelinerun_duration_seconds_count` only have results from cluster `stone-prd-rh01`.
> 
> Two questions:
> 
>     1. Any idea why? Is that ok?

It is OK, and is not atypical @yftacherzog .  In fact, I'm seeing that even from the metrics console and cluster specific grafana panels that appstudio maintains.  In short, the `stone-prd-m01` cluster or our "multi tenant prod cluster" is actually seldom used at this point.  And with the move away from RHTAP providing a service to non-Red Hat users, this is the cluster, along with the its stage counterpart, going away at some point (though I have heard the precise time table for that).

> 
>     2. Would that be the expected result for your query in that case or no entries is a scenario the rule should cover?

No we are not concerned with defining rules over the lack of results from the query.  As long as alerts do not fire on a cluster with no activity, we are good.

rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
gabemontero: Sure I'll break up the line, and sort out what `{{ $value | humanizePercentage }} ` provides exactly and curate the message accordingly @yftacherzog thanks

yftacherzog (approved):
lgtm

gabemontero:
cool thanks for all the help @yftacherzog

PR #108, opened 2023-11-13
chore(DEVHAS-506): Add component alerting rules and runbook links

kim-tsao:
- Add component deletion alerting rules and tests
- ~Update runbook URL in application alerting rules and tests~ will handle in a separate PR


yftacherzog (changes requested):
Can we have the part of updating the runbook url on a PR of its own?

rhobs/alerting/data_plane/prometheus.component_alerts.yaml
```diff
@@ -0,0 +1,28 @@
...
+    - alert: ComponentDeletionErrors
+      expr: |
+        (increase(has_component_failed_deletion_total[1h]))
+        /
+        (increase(has_component_deletion_total[1h]))  > 0.05
+      for: 5m
+      labels:
+        severity: warning
+        slo: true
+      annotations:
```
yftacherzog, line 21: I think we can add the actual query value to the summary/description. You can use `{{ $value }}` for that.

test/promql/tests/data_plane/component_errors_test.yaml
```diff
@@ -0,0 +1,89 @@
+evaluation_interval: 1m
+
+rule_files:
+  - prometheus.component_alerts.yaml
+
+tests:
```
yftacherzog, line 6: Can we have test cases for zero increase on the metrics being used? Especially testing what we get when the denominator is zero and the query returns NaN.

test/promql/tests/data_plane/component_errors_test.yaml
```diff
@@ -0,0 +1,89 @@
...
+              runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/has/component-create-failed.md
+
+  - interval: 1m
+    input_series:
+
+      # HAS experienced ~10% (10.9%) deletion failure rate, should still alert
+      - series: 'has_component_deletion_total{namespace="application-service", pod="has", source_cluster="cluster01"}'
+        values: '10+10x64'
+      - series: 'has_component_failed_deletion_total{namespace="application-service", pod="has", source_cluster="cluster01"}'
+        values: '0x20 0+10x7 70x38'
```
yftacherzog, line 43: The counter is resetting here. It's a good test case.

I think we need one in which the counter does not reset as well.
Also, I think including the query value on the output will make it clear if the reset is being taken into account.

kim-tsao:
> Can we have the part of updating the runbook url on a PR of its own?

Do you mean update the runbook URL for Applications in a separate PR?  The runbook URLs for component can remain in this PR?

test/promql/tests/data_plane/component_errors_test.yaml
kim-tsao: I checked our charts and adjusted the range to see when this could happen.  The only scenario that makes sense is if there's a period of inactivity where there are no deletion requests and no failures

test/promql/tests/data_plane/component_errors_test.yaml
kim-tsao: The counter is not really resetting.  The scenario represents no failures for first 20 mins, then an incrementing series of failures for the next 7 mins until it plateaus.    I'll try to figure out a counter reset scenario

test/promql/tests/data_plane/component_errors_test.yaml
yftacherzog: Is this something that is unlikely to happen?

test/promql/tests/data_plane/component_errors_test.yaml
yftacherzog: You're right. I got confused.

test/promql/tests/data_plane/component_errors_test.yaml
yftacherzog: I see you have this one covered. Cool!

yftacherzog (approved):
lgtm

PR #112, opened 2023-11-15
feat(STONEINTG-561): add alert IS SLO 5

jencull:
add alert for snapshot marked as passed to
release created where >10% of requests takes
longer than 10s
Add alert testing, 3 cases with 2 clusters each:
Where neither cluster triggers alert
Where one cluster triggers alert
Where both clusters trigger alert

### DIFF
--- a/rhobs/alerting/data_plane/prometheus.latency_release_creation_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.latency_release_creation_alerts.yaml
@@ -1,0 +1,33 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: rhtap-latency-release-creation-alerting-rules
+  labels:
+    tenant: rhtap
+spec:
+  groups:
+  - name: latency-release-creation
+    interval: 1m
+    rules:
+    - alert: LatencyReleaseCreation
+      expr: |
+        (
+          (
+            increase(integration_svc_release_latency_seconds_bucket{le="+Inf"}[5m])
+            - ignoring(le)
+            increase(integration_svc_release_latency_seconds_bucket{le="10"}[5m])
+          ) / ignoring(le)
+          increase(integration_svc_release_latency_seconds_bucket{le="+Inf"}[5m])
+        ) > 0.10
+      for: 1m
+      labels:
+        severity: warning
+        slo: true
+      annotations:
+        summary: >-
+          Latency of release creation time exceeded
+        description: >
+          Time from Snapshot marked as passed to release created has been over
+          10s for more than 10% of requests during the last 5 minutes on cluster
+          {{ $labels.source_cluster }}
+        runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/integration-service/latency_release_creation.md
--- a/test/promql/tests/data_plane/latency_release_creation_test.yaml
+++ b/test/promql/tests/data_plane/latency_release_creation_test.yaml
@@ -1,0 +1,99 @@
+evaluation_interval: 1m
+
+rule_files:
+  - 'prometheus.latency_release_creation_alerts.yaml'
+
+tests:
+  - interval: 1m
+    input_series:
+      # Simulating data from Cluster 1 (crosses the 10% threshold)
+      - series: 'integration_svc_release_latency_seconds_bucket{le="10", source_cluster="cluster01"}'
+        values: '0+10x10'  # 10 requests took less than 10s
+      - series: 'integration_svc_release_latency_seconds_bucket{le="+Inf", source_cluster="cluster01"}'
+        values: '0+100x10'  # 100 total occurrences in cluster01
+
+      # Simulating data from Cluster 2 (does not cross the 10% threshold)
+      - series: 'integration_svc_release_latency_seconds_bucket{le="10", source_cluster="cluster02"}'
+        values: '0+90x10'  # 90 requests took less than 10s
+      - series: 'integration_svc_release_latency_seconds_bucket{le="+Inf", source_cluster="cluster02"}'
+        values: '0+100x10'  # 100 total occurrences in cluster02
+
+    alert_rule_test:
+      - eval_time: 10m
+        alertname: LatencyReleaseCreation
+        exp_alerts:
+          - exp_labels:
+              severity: warning
+              slo: true
+              source_cluster: cluster01
+            exp_annotations:
+              summary: Latency of release creation time exceeded
+              description: >
+                Time from Snapshot marked as passed to release created has been over
+                10s for more than 10% of requests during the last 5 minutes on cluster
+                cluster01
+              runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/integration-service/latency_release_creation.md
+
+# Scenario where both clusters cross the 10% threshold
+# Alert triggered for both clusters
+  - interval: 1m
+    input_series:
+      # Simulating data from Cluster 1 (crosses the 10% threshold)
+      - series: 'integration_svc_release_latency_seconds_bucket{le="10", source_cluster="cluster01"}'
+        values: '0+5x10'  # 5 requests took less than 10s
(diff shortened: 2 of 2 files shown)


rhobs/alerting/data_plane/prometheus.latency_release_creation_alerts.yaml
```diff
@@ -0,0 +1,33 @@
...
+         - ignoring(le)
+         increase(release_latency_seconds_bucket{le="10"}[5m])
+        ) / ignoring(le)
+        increase(release_latency_seconds_bucket{le="+Inf"}[5m])
+       ) > 0.10
+      for: 1m
+      labels:
+        severity: warning
+        slo: true
+      annotations:
```
yftacherzog, line 26: I suggest adding the actual query value in the summary/description. {{ $value | humanizePercentage }}

rhobs/alerting/data_plane/prometheus.latency_release_creation_alerts.yaml
```diff
@@ -0,0 +1,33 @@
...
+        summary: >-
+          Snapshot marked as passed to Release created time exceeded
+        description: >
```
yftacherzog, line 29: I see more than one result per cluster. If they are not to be aggregated then I think the description should include the label that sets them apart.

yftacherzog (changes requested):
See my comments. Otherwise, looks good.

rhobs/alerting/data_plane/prometheus.latency_release_creation_alerts.yaml
jencull: done, thank you.

yftacherzog: I don't see it.

jencull: It was uploading as I left you the message. You are too efficient!! 
https://github.com/redhat-appstudio/o11y/pull/112/files

yftacherzog: Consider this query:
```
release_latency_seconds_bucket{source_cluster="stone-prd-rh01", le="10"}
```
If I run it in AppSRE Grafana, I get two results. It means that there's at least one label which set those two apart.
It also means that if your current alert will fire, you won't be able to tell which of those queries had an issue.

Now the question is whether the two needs to be considered together somehow, and then they need to be aggregated (e.g. with `sum`), or each of them should stand on its own, and then we need to add some more labels to the alert description or summary so we'll be able to tell which was is at fault.

jencull: currently on our own grafana dashboard we have just one output to our graphs so we are aggregating. This is our query for the time series graph histogram_quantile(0.90, sum(rate(release_latency_seconds_bucket[$__rate_interval])) by (le)).

yftacherzog: The alert is ultimately going to run on the data series visible in AppSRE Grafana so need to make it applicable there. The query above is also not what we have on the alert rule in this PR.

amisstea: @jencull what was the outcome here?

The `release_latency_seconds_bucket` metric is [collected](https://grafana.app-sre.devshift.net/explore?orgId=1&left=%7B%22datasource%22:%22P22466E8E7855F1E0%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22P22466E8E7855F1E0%22%7D,%22editorMode%22:%22code%22,%22expr%22:%22%20%20%20%20%20%20%20%28%5Cn%20%20%20%20%20%20%20%20%28%5Cn%20%20%20%20%20%20%20%20%20increase%28release_latency_seconds_bucket%7Ble%3D%5C%22%2BInf%5C%22%7D%5B5m%5D%29%5Cn%20%20%20%20%20%20%20%20%20-%20ignoring%28le%29%5Cn%20%20%20%20%20%20%20%20%20increase%28release_latency_seconds_bucket%7Ble%3D%5C%2210%5C%22%7D%5B5m%5D%29%5Cn%20%20%20%20%20%20%20%20%29%20%2F%20ignoring%28le%29%5Cn%20%20%20%20%20%20%20%20increase%28release_latency_seconds_bucket%7Ble%3D%5C%22%2BInf%5C%22%7D%5B5m%5D%29%5Cn%20%20%20%20%20%20%20%29%22,%22legendFormat%22:%22__auto%22,%22range%22:true,%22instant%22:true%7D%5D,%22range%22:%7B%22from%22:%22now-1h%22,%22to%22:%22now%22%7D%7D) for both the integration and release services. Is this expected?

jencull: @yftacherzog I brought this to the team yesterday and the choice was cluster names. I have amended the PR now to reflect this. 

I provided the time series graph SLO query yesterday, apologies. This is our violations panel query: 

sum(rate(release_latency_seconds_bucket{le="10"}[$__rate_interval])) by (job) / sum(rate(release_latency_seconds_count[$__rate_interval])) by (job)*100

however, based on @amisstea comment it now looks like that other metric might be from another team, release service not integration service? if i am understanding that correctly.

All of my team are on PTO today for a public holiday so I will not be able to get clarification on this. It could be that rewriting our metric and changing the name will resolve our issues. I will mark this as a WIP and follow up on Monday, you are agreeable.

yftacherzog: In that case, I suggest changing the query so it only picks your own metrics.

yftacherzog: I can't see the change.

jencull: the change is the prefix integration_svc_ in the expression https://github.com/redhat-appstudio/o11y/pull/112/files We rewrote the metrics to differentiate between us and release service

yftacherzog: oh. I see now.

rhobs/alerting/data_plane/prometheus.latency_release_creation_alerts.yaml
```diff
@@ -0,0 +1,33 @@
...
+    interval: 1m
+    rules:
+    - alert: LatencyReleaseCreation
+      expr: |
+        (
+          sum by (source_cluster) (rate(integration_svc_release_latency_seconds_bucket{le="+Inf"}[5m]))
+          -
+          sum by (source_cluster) (rate(integration_svc_release_latency_seconds_bucket{le="10"}[5m]))
+        ) /
+         sum by (source_cluster) (rate(integration_svc_release_latency_seconds_bucket{le="+Inf"}[5m]))
```
yftacherzog, line 19: Are we expecting more than one time series per cluster for those queries (`integration_svc_release_latency_seconds_bucket{le="X"}`)?
Otherwise, no need to use `sum`.

jencull: the info on the legend is much easier to read with the sum by (source_cluster). So for clarity and presentation for the reader I put it in. 

image

this is what it looks like without 

image

If the presentation doesn't matter so much thats no problem, I'll just remove it. It doesn't change the relevant information, just how it looks on the panel.

yftacherzog: There's no presentation in this context, so it indeed doesn't matter.

jencull: sum removed. Thanks for your guidance.

yftacherzog (approved):
lgtm

PR #114, opened 2023-11-16
chore(RHTAPWATCH-514): limit namespace for control plane alerts

yftacherzog:
The metric currently used for the control plane alerts is also going to be forwarded from the data plane clusters to allow some other alerts.

With this change, we specify the exact namespace for the control plane alerts so they don't trigger on metrics coming from other namespaces.

### DIFF
--- a/rhobs/alerting/control_plane/prometheus.argocd_alerts.yaml
+++ b/rhobs/alerting/control_plane/prometheus.argocd_alerts.yaml
@@ -10,7 +10,10 @@ spec:
     interval: 1m
     rules:
     - alert: DegradedArgocdApp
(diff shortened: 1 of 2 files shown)


avi-biton (commented):
Are we sure these are the only namespaces we should query?
If yes, lgtm

yftacherzog:
> Are we sure these are the only namespaces we should query? If yes, lgtm

Yes. It seems as all time series in prod have this label set to argocd and in stage they all set to argocd-staging

avi-biton (approved):
lgtm

amisstea (approved):

PR #115, opened 2023-11-16
chore(PLNSRVCE-1470): adjust pipeline service grafana panels to switch from 'sum_over_time' to 'increase' to match alerts

gabemontero:
During the definition of the accompanying pipeline service alerts we settled on using 'increase' instead of 'sum_over_time' to better deal with metric controller restarts.  Adjusting ccompanying grafana panels accordingly.

@redhat-appstudio/o11y PTAL


amisstea (approved):
LGTM :+1:

@hmariset @Kousalya1998 PTAL

image
_Verified by pasting the `panels` JSON into a copy of the dashboard in the AppSRE stage grafana UI._

PR #117, opened 2023-11-17
feat(GITOPSRVCE-753): Add alert rule/test for Cluster Agent Operation SLO

Rizwana777:
JIRA - https://issues.redhat.com/browse/GITOPSRVCE-753

### DIFF
--- a/rhobs/alerting/data_plane/prometheus.gitops_cluster_agent_op_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.gitops_cluster_agent_op_alerts.yaml
@@ -1,0 +1,24 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: rhtap-gitops-cluster-agent-operations-alerting
+  labels:
+    tenant: rhtap
+spec:
+  groups:
+  - name: cluster-agent-operations_alerts
+    interval: 1m
+    rules:
+    - alert: GitOpsClusterAgentOperationErrors
+      expr: |
+        sum(operationDB_rows_in_non_complete_state) <= 10
+      for: 5m
+      labels:
+        severity: warning
+        slo: true
+      annotations:
+        summary: >-
+          Cluster-Agent Operations in non-complete state are too high
+        description: >-
+          The sum of cluster-agent operations in non-complete state is greater than 10
+        runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/gitops/cluster-agent-operations.md
--- a/rhobs/alerting/data_plane/prometheus.gitops_cluster_agent_op_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.gitops_cluster_agent_op_alerts.yaml
@@ -11,14 +11,14 @@ spec:
     rules:
     - alert: GitOpsClusterAgentOperationErrors
       expr: |
-        sum(operationDB_rows_in_non_complete_state) <= 10
+        operationDB_rows_in_non_complete_state >= 10
       for: 5m
       labels:
         severity: warning
         slo: true
       annotations:
         summary: >-
-          Cluster-Agent Operations in non-complete state are too high
+          Cluster-Agent Operations in non-complete state are too high. Got: {{ $value }}
         description: >-
-          The sum of cluster-agent operations in non-complete state is greater than 10
+          The sum of cluster-agent operations in non-complete state is greater than 10 on cluster {{ $labels.source_cluster }}
         runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/gitops/cluster-agent-operations.md
--- a/test/promql/tests/data_plane/gitops_cluster_agent_operation_test.yaml
+++ b/test/promql/tests/data_plane/gitops_cluster_agent_operation_test.yaml
@@ -1,0 +1,57 @@
+evaluation_interval: 1m
+
+rule_files:
+  - prometheus.gitops_cluster_agent_op_alerts.yaml
+
+tests:
+  # ----- Cluster-Agent operations in non-complete state -----
+  - interval: 1m
+    input_series:
+      # More than 10 cluster-agent operations are in non-complete state, should alert
+      - series: 'operationDB_rows_in_non_complete_state'
+        values: '1x11'   # 11 cluster-agent operations are in non-complete state
+
+    alert_rule_test:
+      - eval_time: 11m
+        alertname: GitOpsClusterAgentOperationErrors
+        exp_alerts:
+          - exp_labels:
+              severity: warning
+              slo: true
+            exp_annotations:
+              summary: >-
+                Cluster-Agent Operations in non-complete state are too high
+              description: >-
+                The sum of cluster-agent operations in non-complete state is greater than 10
+              runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/gitops/cluster-agent-operations.md
+
+  - interval: 1m
+    input_series:
+      # 10 cluster-agent operations are in non-complete state, should alert
+      - series: 'operationDB_rows_in_non_complete_state'
+        values: '1x10'   # 10 cluster-agent operations are in non-complete state
+
+    alert_rule_test:
+      - eval_time: 10m
+        alertname: GitOpsClusterAgentOperationErrors
+        exp_alerts:
+          - exp_labels:
+              severity: warning
+              slo: true
+            exp_annotations:
+              summary: >-
+                Cluster-Agent Operations in non-complete state are too high
+              description: >-
+                The sum of cluster-agent operations in non-complete state is greater than 10
+              runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/gitops/cluster-agent-operations.md
+
+  - interval: 1m
+    input_series:
+
+      # Less than 10 cluster-agent operations are in non-complete state, should not alert
+      - series: 'operationDB_rows_in_non_complete_state'
+        values: '1x3'   # 3 cluster-agent operations are in non-complete state
+
+    alert_rule_test:
+      - eval_time: 3m
+        alertname: GitOpsClusterAgentOperationErrors
--- a/test/promql/tests/data_plane/gitops_cluster_agent_operation_test.yaml
+++ b/test/promql/tests/data_plane/gitops_cluster_agent_operation_test.yaml
@@ -8,50 +8,52 @@ tests:
   - interval: 1m
     input_series:
       # More than 10 cluster-agent operations are in non-complete state, should alert
-      - series: 'operationDB_rows_in_non_complete_state'
-        values: '1x11'   # 11 cluster-agent operations are in non-complete state
+      - series: 'operationDB_rows_in_non_complete_state{source_cluster="cluster01"}'
+        values: '11x11'   # 11 cluster-agent operations are in non-complete state
 
     alert_rule_test:
-      - eval_time: 11m
+      - eval_time: 6m
         alertname: GitOpsClusterAgentOperationErrors
         exp_alerts:
           - exp_labels:
               severity: warning
               slo: true
+              source_cluster: cluster01
             exp_annotations:
               summary: >-
-                Cluster-Agent Operations in non-complete state are too high
+                Cluster-Agent Operations in non-complete state are too high. Got: 11
               description: >-
-                The sum of cluster-agent operations in non-complete state is greater than 10
+                The sum of cluster-agent operations in non-complete state is greater than 10 on cluster cluster01
               runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/gitops/cluster-agent-operations.md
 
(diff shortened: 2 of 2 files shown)


rhobs/alerting/data_plane/prometheus.gitops_cluster_agent_op_alerts.yaml
```diff
@@ -0,0 +1,24 @@
...
+  labels:
+    tenant: rhtap
+spec:
+  groups:
+  - name: cluster-agent-operations_alerts
+    interval: 1m
+    rules:
+    - alert: GitOpsClusterAgentOperationErrors
+      expr: |
+        sum(operationDB_rows_in_non_complete_state) <= 10
```
keithchong, line 14: Question regarding this metric and also in the grafana dashboard visualization.  Is this metric already the sum total of the number of rows in non-complete state?  Comparing this to, for example, the metric `operationDB_rows_in_completed_state` which returns a positive integer number >=0.  In the sandbox member cluster, this number currently returns 288.

keithchong, line 14: Should it be >= 10 or maybe from the Jira, >10 ?  eg. ......report when it exceeds a given value (e.g. 10).

test/promql/tests/data_plane/gitops_cluster_agent_operation_test.yaml
```diff
@@ -0,0 +1,57 @@
...
+rule_files:
+  - prometheus.gitops_cluster_agent_op_alerts.yaml
+
+tests:
+  # ----- Cluster-Agent operations in non-complete state -----
+  - interval: 1m
+    input_series:
+      # More than 10 cluster-agent operations are in non-complete state, should alert
+      - series: 'operationDB_rows_in_non_complete_state'
+        values: '1x11'   # 11 cluster-agent operations are in non-complete state
```
keithchong, line 12: Make this `11x11`, if we use `>10`

test/promql/tests/data_plane/gitops_cluster_agent_operation_test.yaml
```diff
@@ -0,0 +1,57 @@
...
+                Cluster-Agent Operations in non-complete state are too high
+              description: >-
+                The sum of cluster-agent operations in non-complete state is greater than 10
+              runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/gitops/cluster-agent-operations.md
+
+  - interval: 1m
+    input_series:
+      # 10 cluster-agent operations are in non-complete state, should alert
+      - series: 'operationDB_rows_in_non_complete_state'
+        values: '1x10'   # 10 cluster-agent operations are in non-complete state
```
keithchong, line 32: This probably should be `10x10` 

If you want to use `>10`, then we should not get any alerts for this test. If you want the alert, then use `11x10` instead

rhobs/alerting/data_plane/prometheus.gitops_cluster_agent_op_alerts.yaml
yftacherzog, line 14: Do we need to use `sum` here? Looking at AppSRE Grafana, I see a single data series per cluster for this metric.
Using sum here means that we're summing over multiple clusters, which I'm not sure is the intention.

rhobs/alerting/data_plane/prometheus.gitops_cluster_agent_op_alerts.yaml
```diff
@@ -0,0 +1,24 @@
...
+      for: 5m
+      labels:
+        severity: warning
+        slo: true
+      annotations:
+        summary: >-
+          Cluster-Agent Operations in non-complete state are too high
```
yftacherzog, line 21: I suggest including `{{ $value }}` in the summary/description to have better visibility of the query result.

test/promql/tests/data_plane/gitops_cluster_agent_operation_test.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+  - interval: 1m
+    input_series:
+
+      # Less than 10 cluster-agent operations are in non-complete state, should not alert
+      - series: 'operationDB_rows_in_non_complete_state'
+        values: '1x3'   # 3 cluster-agent operations are in non-complete state
+
+    alert_rule_test:
+      - eval_time: 3m
```
yftacherzog, line 56: The reason this one is not alerting is because the rule is waiting for 5 minutes before an alert is fired, but this test case is only waiting 3 minutes.

yftacherzog (changes requested):
Few points:

1. Please verify whether the alert should be fired when the value is lower than 10 or higher than 10, per @keithchong's comment.
2. Please check about the use of `sum` within the query. I'm not convinced it should be used, and if it does, I suspect we should not sum across clusters.
Note that `sum` is not summarizing over time. It summarizes the value at a given moment for all **time series** that fit the query. In this case, I think there might be only one time series per cluster, so I don't think `sum` should be used.
3. In Prometheus  tests cases, each entry on the `input_series` list is comprised of a metric along with its labels (`series`) and its values over time (`values`).
The values are provided as a string of space-separated numbers, each representing the value of the series at a given time, where the time increments between values is the `interval` value.

```
      - series: 'operationDB_rows_in_non_complete_state'
        values: '1x10'
```
is a shortened notation for value of `1` over 10 intervals. i.e.
```
      - series: 'operationDB_rows_in_non_complete_state'
        values: '1 1 1 1 1 1 1 1 1 1'
```
With `interval` of `1m`, this means 10 occurrences of value `1` over ten minutes for this metric.

Also, the test will examine the rule at `eval_time`. In this case, it will find there `1` and the query will sum it with all time series for the 'operationDB_rows_in_non_complete_state', and in this case, there's only one such time series (only one element on the `input_series` list), so the result will still be `1`.

The different test cases here should not differ by the number of entries, nor the evaluation time. They should differ by the values over time for the time series.

Reference: https://prometheus.io/docs/prometheus/latest/configuration/unit_testing_rules/

rhobs/alerting/data_plane/prometheus.gitops_cluster_agent_op_alerts.yaml
Rizwana777: yes @keithchong The metric `operationDB_rows_in_non_complete_state`  returns the total number of operation db rows in non-complete state, if we take an example of operationDB_rows_in_completed_state , in  prod multitenant grafana dashboard, its value is 144, In the multitenant grafana  dashboard we are checking the values from last 1 hour, so it will return the total operationDB_rows_in_completed_state  from last 1 hour i.e 144, similar case is with operationDB_rows_in_non_complete_state it will return total number of  operation db rows in non-complete state from last 1 hour, as time-range is set for 1 hour, but in RHTAP SLO dashboard the time range is set for 28d so it return multiple values i.e, it will return all the values from past 28d like this 
Screenshot 2023-11-20 at 12 52 29 PM

So I wanted to sum up all the `operationDB_rows_in_non_complete_state` values from past 28d

Rizwana777: Is there any other way to aggregate the data from past 28d based on time range or we can add the time limit by ourselves like this, `sum(increase(operationDB_rows_in_non_complete_state[24h]))`, please let me know any other possible ways to do this

Rizwana777: Screenshot 2023-11-20 at 4 22 06 PM
basically what I am doing is if  $A <= 10 , if 3 <= 10 i.e true (then no alert), if 12 <= 10 i.e, false (alert), please let me know whether I should be doing this other way around?

Rizwana777: I have added a comment above regarding this https://github.com/redhat-appstudio/o11y/pull/117#discussion_r1399010413 please let me know your suggestions as well

Rizwana777:
> Few points:
> 
> 1. Please verify whether the alert should be fired when the value is lower than 10 or higher than 10, per @keithchong's comment.
> 2. Please check about the use of `sum` within the query. I'm not convinced it should be used, and if it does, I suspect we should not sum across clusters.
>    Note that `sum` is not summarizing over time. It summarizes the value at a given moment for all **time series** that fit the query. In this case, I think there might be only one time series per cluster, so I don't think `sum` should be used.
> 3. In Prometheus  tests cases, each entry on the `input_series` list is comprised of a metric along with its labels (`series`) and its values over time (`values`).
>    The values are provided as a string of space-separated numbers, each representing the value of the series at a given time, where the time increments between values is the `interval` value.
> 
> ```
>       - series: 'operationDB_rows_in_non_complete_state'
>         values: '1x10'
> ```
> 
> is a shortened notation for value of `1` over 10 intervals. i.e.
> 
> ```
>       - series: 'operationDB_rows_in_non_complete_state'
>         values: '1 1 1 1 1 1 1 1 1 1'
> ```
> 
> With `interval` of `1m`, this means 10 occurrences of value `1` over ten minutes for this metric.
> 
> Also, the test will examine the rule at `eval_time`. In this case, it will find there `1` and the query will sum it with all time series for the 'operationDB_rows_in_non_complete_state', and in this case, there's only one such time series (only one element on the `input_series` list), so the result will still be `1`.
> 
> The different test cases here should not differ by the number of entries, nor the evaluation time. They should differ by the values over time for the time series.
> 
> Reference: https://prometheus.io/docs/prometheus/latest/configuration/unit_testing_rules/

 I have added my comments above for first 2  points, regarding lower than or higher than and regarding sum, once first 2 points gets cleared, I will look into the test comments, thank you

rhobs/alerting/data_plane/prometheus.gitops_cluster_agent_op_alerts.yaml
yftacherzog: Is state complete means active or state non-complete means active?

I'm not sure what are we trying to check. What's the reason for aggregating over time?
If I understand correctly, this metric is a gauge. What's the meaning of summing its value over time?

Rizwana777: I think I should change this to  >=10

yftacherzog: If it should be always less than 10, do we need to sum over time? The requirement seem to be about the current value. I don't see it implemented by summing over time.

Rizwana777: And in ticket it is mentioned as <=10 cluster-agent operations should b active `at any given time`

yftacherzog: The alert is being evaluated every minute. At the time of evaluation, the alert query is run and either returns some time series or not.
Depending on the value given for the `for` field, a time series need to be returned in multiple consecutive such evaluations before the alert will fire. The `for` field is there to eliminate noise.

With that in mind, I think this query needs to be without the sum, and the `for` value should be set according to what you see fit.

Rizwana777:
Hi @keithchong and @yftacherzog updated all the changes and updated dashboard as well and here is the [PR](https://github.com/redhat-appstudio/o11y/pull/123), PTAL

yftacherzog (dismissed):

test/promql/tests/data_plane/gitops_cluster_agent_operation_test.yaml
keithchong: I see your change, looks good.

test/promql/tests/data_plane/gitops_cluster_agent_operation_test.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+  - interval: 1m
+    input_series:
+
+      # Less than 10 cluster-agent operations are in non-complete state, should not alert
+      - series: 'operationDB_rows_in_non_complete_state{source_cluster="cluster01"}'
+        values: '5x5'   # 5 cluster-agent operations are in non-complete state
+
+    alert_rule_test:
+      - eval_time: 15m
```
keithchong, line 58: The series is:   5 5 5 5 5
but the eval_time is 15m, although this test does pass, is this the intention of the test? @yftacherzog, should we provide a time series that exceeds the `eval_time` and the 'for' duration? (Just looking at your original review comment for this specific test)

keithchong (commented):
I think it looks good. I just added one last comment/question.

test/promql/tests/data_plane/gitops_cluster_agent_operation_test.yaml
yftacherzog: Yes. We should have the time series at least as long as the eval_time. If we want to test missing-data scenarios, then we should have `_` "measurements" instead of numbers for where the data is missing.

keithchong (approved):
LGTM

Rizwana777:
@yftacherzog updated the changes , can you please take a look as well?

yftacherzog (approved):

PR #119, opened 2023-11-20
chore(RHTAPWATCH-565): Modify the service into Prometheus exporter

Kousalya1998:
- Add exporter created with summy metrics
- Add the unit test file


exporters/dsPrometheusExporter/dsPrometheusExporter.go
yftacherzog, line 1: I see we're creating new files. I don't think we should. Need to change the existing files instead.
I also think we should not use camel case in file names.

exporters/dsPrometheusExporter/dsPrometheusExporter.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+func ping(w http.ResponseWriter, req *http.Request) {
+	if req.URL.Path != "/ping" {
+		// This handles unknown endpoints like 404 status code and an empty response body
+		w.WriteHeader(http.StatusNotFound)
+		return
+	}
+
+	pingCounter.Inc()
```
yftacherzog, line 25: The code as implemented here is demonstrating a self-instrumenting service.
In other words, the code is generating metrics along the way, rather than collecting them when `/metrics` is accessed.

The datasource exporter is going to export metric from another system (the grafana operator) upon accessing the `/metrics` endpoint.

I think that if we implement this sort of dummy exporter now, we will have a big gap between this stage and the one in which we implement the actual exporter logic.

I suggest that we instead implement now a dummy exporter that is more on the path between the dummy service and the intended exporter. 

Some background here: https://prometheus.io/docs/instrumenting/writing_exporters/

There's an interface that such exporter needs to implement. It is explained here: https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#Collector

You can also use the dummy exporter on the ticket as an example.

exporters/dsPrometheusExporter/dsPrometheusExporter.go
```diff
@@ -0,0 +1,35 @@
...
+	"net/http"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+var pingCounter = prometheus.NewCounter(
+	prometheus.CounterOpts{
+		Name: "ping_request_count",
+		Help: "No of request handled by Ping handler",
```
avi-biton, line 14:
```suggestion
		Help: "Number of requests handled by Ping handler",
```

exporters/dsPrometheusExporter/dsPrometheusExporter_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPingHandler(t *testing.T) {
+	// Reset the pingCounter before each test
+	prometheus.Unregister(pingCounter)
+	pingCounter = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Name: "ping_request_count",
+			Help: "No of requests handled by Ping handler",
```
avi-biton, line 19:
```suggestion
			Help: "Number of requests handled by Ping handler",
```

exporters/dsPrometheusExporter/dsPrometheusExporter.go
Kousalya1998: @yftacherzog 
My understanding is: that you recommending a different approach where we create a separate dummy exporter that collects and exports metrics from the dummy service.

Something like a custom Prometheus collector like the `fooCollector` (one you used in your [example code](https://github.com/yftacherzog/prometheus-exporter))

yftacherzog: The _service_ in the context of this epic _is_ the exporter. It's ultimately going to export metrics it will generate for the Grafana datasource.

We started by creating it as a dummy service in order to be able to create it in as smaller increments as possible. i.e. we create something that can be deployed in Kubernetes -> we make it behave like an exporter -> we make it export Grafana datasource metrics.

All in all, we only create one service -- the exporter.
We're at the point of turning it from a dummy service to a dummy exporter (this ticket and PR).
Next, we'll make it export meaningful stuff.

exporters/dsexporter/dsexporter.go
```diff
@@ -3,18 +3,33 @@ package main
...
 func main() {
-	http.HandleFunc("/ping", ping)
+	prometheus.MustRegister(pingCounter)
 
-	fmt.Println("Starting dummy service on :8090")
-	err := http.ListenAndServe(":8090", nil)
-	if err != nil {
-		fmt.Println("Error starting dummy service:", err)
-	}
+	http.HandleFunc("/ping", ping)
```
yftacherzog, line 32: There shouldn't be a ping endpoint at this stage. We should now only listen to the metrics endpoint.

exporters/dsPrometheusExporter/dsPrometheusExporter.go
Kousalya1998: @yftacherzog That makes sense. Thank you for providing a detailed explanation of the ticket's goal.

exporters/dsexporter/dsexporter.go
```diff
@@ -1,20 +1,46 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
-func ping(w http.ResponseWriter, req *http.Request) {
-	fmt.Fprintf(w, "pong")
+// CustomExporter is a dummy exporter that implements the prometheus.Collector interface.
+// The Custom exporter is a way to structure our code so that it can generate and expose metrics.
+type CustomExporter struct {
+	pingCounter prometheus.Counter
```
yftacherzog, line 13: With the current implementation, I think calling it `emptyCounter` is a better representation of what it does.
In any case, we should not have the word `ping` on the exporter-related files after this PR.

exporters/dsexporter/dsexporter.go
```diff
@@ -1,20 +1,46 @@
...
+// After Prometheus understands what metrics are available (using the `Describe` method),
+// it then calls the `Collect` method to actually get the values of those metrics.
+func (e *CustomExporter) Collect(ch chan<- prometheus.Metric) {
+	e.pingCounter.Collect(ch)
 }
 
 func main() {
-	http.HandleFunc("/ping", ping)
+	exporter := NewCustomExporter()
+	prometheus.MustRegister(exporter)
```
avi-biton, line 42:
```suggestion
        reg := prometheus.NewPedanticRegistry()
	exporter := NewCustomExporter()
	reg.MustRegister(exporter)

	// http.Handle("/metrics", promhttp.Handler())
	// http.ListenAndServe(":8090", nil)
	http.Handle("/metrics", promhttp.HandlerFor(
		reg,
		promhttp.HandlerOpts{
			EnableOpenMetrics: true,
			Registry:          reg,
		},
	))
	http.ListenAndServe(":8090", nil)
```

When we are using the default registry, we get a lot of predefined metrics that we do not need (take a look at the /metrics endpoint).
We can instead use [NewPedanticRegistry](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#NewPedanticRegistry) to get a clean /metrics with only our metric

exporters/dsexporter/dsexporter.go
Kousalya1998: @yftacherzog What do you mean by exporter-related files here?
quick question: since our focus should be on transitioning from a dummy service to a dummy exporter, I named it as `CustomExporter`

exporters/dsexporter/dsexporter.go
```diff
@@ -1,20 +1,46 @@
...
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
-func ping(w http.ResponseWriter, req *http.Request) {
-	fmt.Fprintf(w, "pong")
+// CustomExporter is a dummy exporter that implements the prometheus.Collector interface.
+// The Custom exporter is a way to structure our code so that it can generate and expose metrics.
+type CustomExporter struct {
```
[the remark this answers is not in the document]
Kousalya1998: @avi-biton quick question: Since our focus should be on transitioning from a dummy service to a dummy exporter, I named it to be CustomExporter. Correct me if I am wrong.

exporters/dsexporter/dsexporter.go
Kousalya1998: Yes, I will take a look at the [NewPedanticRegistry](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#NewPedanticRegistry) to get a clean /metrics with only our metric

exporters/dsexporter/dsexporter.go
yftacherzog: I agree with @avi-biton about naming the type `CustomCollector` as it implements the `Collector` interface (https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#Collector).

yftacherzog (changes requested):
`ping` is still mentioned in multiple places. Let's un-ping this.

exporters/dsexporter/dsexporter.go
Kousalya1998: Okay renamed it to `CustomCollector` and removed the `ping` keywords

yftacherzog (approved):
lgtm.
Don't forget to push the new image once this is merged - as per the acceptance criteria.

PR #131, opened 2023-12-06
feat(STONEINTG-559): add alert and test for SLO 6 Intg Svc

jencull:
add alert rule and testing for same for SLO 6
Integration Service - Snapshot Environment Binding created to ready seconds

### DIFF
--- a/rhobs/alerting/data_plane/prometheus.seb_created_to_ready_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.seb_created_to_ready_alerts.yaml
@@ -1,0 +1,33 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: rhtap-seb-created-to-ready-alerting-rules
+  labels:
+    tenant: rhtap
+spec:
+  groups:
+  - name: seb-created-to-ready
+    interval: 1m
+    rules:
+    - alert: SebCreatedToReady
+      expr: |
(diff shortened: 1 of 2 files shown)


jencull:
This metric gets 1 hit every few days so 'no data' is the usual return on our own grafana panels. We had one measurement late in the day on 5th Dec. If 'no data' is returned that is to be expected. Screenshots of today (Dec 6th) panels, for reference:

image

image

yftacherzog:
> This metric gets 1 hit every few days so 'no data' is the usual return on our own grafana panels. We had one measurement late in the day on 5th Dec. If 'no data' is returned that is to be expected. Screenshots of today (Dec 6th) panels, for reference:
> 
> image
> 
> image

Do we have a test case for that common case? i.e. confirm that things work as expected when both metrics are at zero throughout the test.

jencull:
> > This metric gets 1 hit every few days so 'no data' is the usual return on our own grafana panels. We had one measurement late in the day on 5th Dec. If 'no data' is returned that is to be expected. Screenshots of today (Dec 6th) panels, for reference:
> > image
> > image
> 
> Do we have a test case for that common case? i.e. confirm that things work as expected when both metrics are at zero throughout the test.

There will be, later today. Thanks :)

jencull:
That test is added now.

yftacherzog (approved):
lgtm

amisstea (approved):

PR #136, opened 2023-12-13
feat(GITOPSRVCE-751): Add alert rule/test for Route SLO

jparsai:
## Description
This PR is to add alert rules and their tests for Routes SLOs added by GitOps Service team.

## Jira Ticket
https://issues.redhat.com/browse/GITOPSRVCE-751

### DIFF
--- a/rhobs/alerting/data_plane/prometheus.gitops_routes_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.gitops_routes_alerts.yaml
@@ -1,0 +1,32 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: rhtap-gitops-routes-alerting
+  labels:
+    tenant: rhtap
+spec:
+  groups:
+  - name: gitops_routes_alerts
+    interval: 1m
+    rules:
+    - alert: GitOpsRoutesHealthErrors
+      expr: |
+        (
+          sum(openshift_route_status{ status="True", namespace="gitops-service-argocd"})
+          by (source_cluster)
+        )
+        /
+        (
+          sum(openshift_route_status{namespace="gitops-service-argocd"})
+          by (source_cluster)
+        ) < 0.95
+      for: 5m
+      labels:
+        severity: warning
+        slo: "true"
+      annotations:
+        summary: >-
+          Less than 95% of GitOps Routes are in Healthy state: {{ $value | humanizePercentage }} in cluster: {{ $labels.source_cluster }}
+        description: >-
+          The percentage total of Argo CD Routes that are in Healthy state is less than 95% in cluster: {{ $labels.source_cluster }}.
+        runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/gitops/routes.md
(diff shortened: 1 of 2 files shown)


test/promql/tests/data_plane/gitops_routes_test.yaml
```diff
@@ -0,0 +1,408 @@
...
+  - prometheus.gitops_routes_alerts.yaml
+
+tests:
+  - interval: 1m
+    input_series:
+      - series: 'openshift_route_status{status="True", name="route-001", namespace="gitops-service-argocd"}'
+        values: '0 0 0 0 0 0 0 0 0 0 0'
+
+      - series: 'openshift_route_status{status="False",  name="route-002", namespace="gitops-service-argocd"}'
+        values: '1 1 1 1 1 1 1 1 1 1 1'
```
yftacherzog, line 13: Does the `openshift_route_status` metric gives `0` on the non-current status, or does it give no data?
e.g. when `openshift_route_status{status="False"}` is at `1`, would `openshift_route_status{status="True"}` would be on `0` or it will just not provide any data?

jparsai: If `openshift_route_status{status="False"}` is at 1 `openshift_route_status{status="True"}` will have no row.

yftacherzog: If I understand correctly, that metric cannot have a value of zero. If that is the case, I suggest dropping the test cases having that value, as they might be misleading for the behavior of the metric.

yftacherzog: The test case that I think we're missing is the one in which we have a single route failing. It should be something like that:
```
      - series: 'openshift_route_status{status="True", name="route-001", namespace="gitops-service-argocd"}'
        values: '_ _ _ _ _ _ _ _ _ _ _ '

      - series: 'openshift_route_status{status="False",  name="route-001", namespace="gitops-service-argocd"}'
```
I think that one might not be working, as the label-true query will not return any time series and no calculation will be performed (and no alert).

This sort of metric is tricky to do calculations on. We had long discussions on a metric with similar behavior here: https://github.com/redhat-appstudio/o11y/pull/94

Have a look to see if it helps you.

jparsai: Yes initially I tried giving value `"_ _ _ _ _"` but as you said it fails so I used 0, but it seems that changes the scenario.

rhobs/alerting/data_plane/prometheus.gitops_routes_alerts.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: rhtap-gitops-routes-alerting
+  labels:
+    tenant: rhtap
+spec:
+  groups:
+  - name: gitops_routess_alerts
```
amisstea, line 9:
```suggestion
  - name: gitops_routes_alerts
```

amisstea (approved):
Only one small typo to correct but lgtm otherwise :+1:

rhobs/alerting/data_plane/prometheus.gitops_routes_alerts.yaml
jparsai: Updated it, PTAL.

PR #137, opened 2023-12-13
feat(GITOPSRVCE-751): Add alert rule/test for StatefulSets SLO

jparsai:
## Description
This PR is to add alert rules and their tests for StatefulSets SLOs added by GitOps Service team.

## Jira Ticket
https://issues.redhat.com/browse/GITOPSRVCE-751

### DIFF
--- a/rhobs/alerting/data_plane/prometheus.gitops_statefulsets_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.gitops_statefulsets_alerts.yaml
@@ -1,0 +1,32 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: rhtap-gitops-statefulsets-alerting
+  labels:
+    tenant: rhtap
+spec:
+  groups:
+  - name: gitops_statefulsets_alerts
+    interval: 1m
+    rules:
+    - alert: GitOpsStatefulSetsHealthErrors
+      expr: |
+        (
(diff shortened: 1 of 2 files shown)


rhobs/alerting/data_plane/prometheus.gitops_statefulsets_alerts.yaml
```diff
@@ -0,0 +1,32 @@
...
+        (
+          sum(kube_statefulset_replicas{namespace="gitops-service-argocd"})
+          by (source_cluster)
+        ) < 0.95
+      for: 5m
+      labels:
+        severity: warning
+        slo: true
+      annotations:
+        summary: >-
```
yftacherzog, line 28: I suggest adding the source_cluster to the summary/description

jparsai: Does it mean that I always need to have source_cluster in all the tests?

yftacherzog: I think it will be empty otherwise. It might be a little burden to set it in the test, but it will make the alert more useful. Otherwise, you won't have a quick way to tell which cluster is alerting.

jparsai: I added it in all test cases, as you said it makes alert more useful, so better to add. PTAL.

yftacherzog (approved):
lgtm

amisstea (approved):

PR #147, opened 2023-12-22
chore: Enable custom build pipeline

hmariset:
Remove .tekton folder to enable custom build pipeline and related PR to get merged.


yftacherzog (changes requested):
I'd expect to see the tests on the old pipelines being included on the new pipelines. Otherwise, we're not running our tests anymore.

hmariset:
> I'd expect to see the tests on the old pipelines being included on the new pipelines. Otherwise, we're not running our tests anymore.

@yftacherzog I have kept the old pull-requests file now to run all the tasks in that and all the checks are also running now. Please review again.

yftacherzog:
> > I'd expect to see the tests on the old pipelines being included on the new pipelines. Otherwise, we're not running our tests anymore.
> 
> @yftacherzog I have kept the old pull-requests file now to run all the tasks in that and all the checks are also running now. Please review again.

What about the content of the push pipeline? Is it covered by the new pipeline?

hmariset:
> > > I'd expect to see the tests on the old pipelines being included on the new pipelines. Otherwise, we're not running our tests anymore.
> > 
> > 
> > @yftacherzog I have kept the old pull-requests file now to run all the tasks in that and all the checks are also running now. Please review again.
> 
> What about the content of the push pipeline? Is it covered by the new pipeline?

@yftacherzog I just included the task from the old push pipeline to the new one.

yftacherzog:
> > > > I'd expect to see the tests on the old pipelines being included on the new pipelines. Otherwise, we're not running our tests anymore.
> > > 
> > > 
> > > @yftacherzog I have kept the old pull-requests file now to run all the tasks in that and all the checks are also running now. Please review again.
> > 
> > 
> > What about the content of the push pipeline? Is it covered by the new pipeline?
> 
> @yftacherzog I just included the task from the old push pipeline to the new one.

Cool. Can you rebase so we get the tests running with the latest changes? I want to make sure that no test failures were introduced when CI was off.

yftacherzog (approved):
lgtm

yftacherzog:
@amisstea can you have a look?

amisstea (changes requested):
We have two pipelines running for pull request events: `on-pull-request` and `o11y-on-pull-request`. We should ideally only have the latter with all important tasks incorporated from the former. Based on a side conversation with @hmariset it appears there are challenges getting the make commands to work. I'd like to see if we can track down the root of that problem before resorting to the current multiple pipeline workaround.

yftacherzog (changes requested):
We don't have some of our steps now.
See for example step `check-and-test-prometheus-rules` on the pull-request yaml being deleted on this PR.

hmariset:
> We don't have some of our steps now. See for example step `check-and-test-prometheus-rules` on the pull-request yaml being deleted on this PR.

I deleted that for testing purpose, when we merge the old pipeline all the old tests are failing due to clone-repository being using different image.

yftacherzog:
> > We don't have some of our steps now. See for example step `check-and-test-prometheus-rules` on the pull-request yaml being deleted on this PR.
> 
> I deleted that for testing purpose, when we merge the old pipeline all the old tests are failing due to clone-repository being using different image.

You mean that the way we clone the repo for the o11y tests and the way the it's cloned by the new pipeline are not compatible with one another? i.e. if you use one of them to clone the repo, the tests of the other will fail? This might be about the directory in which the content is pulled into, but can also be about the way it's pulled (e.g. depth).

If so, is it true both ways? i.e. new clone fails old tests and old clone fails new tests?

@amisstea do you see the part of having everything in one pipelinerun a must-have for now?
What are the disadvantages of running to parallel pipelineruns here?

In any case, if we want to have it on a single pipeline, we can start by having the two clone tasks storing the code to a different workspaces and have the two "streams" of tests running in parallel on different workspaces., and we can later find the issue and optimize it.

amisstea:
> @amisstea do you see the part of having everything in one pipelinerun a must-have for now?

It's not a must-have but we should try to identify the root problem before deciding if this is tech debt worth kicking down the road. It sounds like there may have been a small adjustment to the way the `clone-repository` checks out the code but that hasn't been confirmed AFAIK. @hmariset have you identified what changed in the task yet?

> What are the disadvantages of running to parallel pipelineruns here?

Maybe not much but it could be harder to fix this later as more contributions are made to the pipelinerun definitions.

hmariset:
@yftacherzog Yes, they are not compatible. If I use old clone-repository for new pipeline the tasks are failing and same thing for old pipeline tasks as well. Yeah I too think the problem is with workspace/directory. 
@amisstea Both clone-repository tasks are using same git-clone but the bundle images are different.

hmariset:
@amisstea @yftacherzog The problem is with the workingDir. When I modified the workingDir to `/workspace/workspace/source`  all the old tasks are passing with new clone-repository.

.tekton/o11y-push.yaml
```diff
@@ -0,0 +1,413 @@
...
+      params:
+      - name: ORIGIN_REPO
+        value: $(params.git-url)
+      - name: REVISION
+        value: $(params.revision)
+      - name: SCRIPT
+        value: $(params.infra-deployment-update-script)
+      taskRef:
+        bundle: quay.io/redhat-appstudio-tekton-catalog/task-update-infra-deployments:0.1
+        name: update-infra-deployments
```
amisstea, line 222: This `infra-deployments-mr` task will currently always execute regardless of the status of the other tasks. Shouldn't this be gated (using `runAfter`) by the success of the other post-build checks (e.g. `clamav-scan`, `sbom-json-check`, etc.)? I'm thinking we shouldn't open the PR if there are any failures in this pipeline.

.tekton/o11y-pull-request.yaml
```diff
@@ -0,0 +1,492 @@
...
+kind: PipelineRun
+metadata:
+  annotations:
+    build.appstudio.openshift.io/repo: https://github.com/redhat-appstudio/o11y?rev={{revision}}
+    build.appstudio.redhat.com/commit_sha: '{{revision}}'
+    build.appstudio.redhat.com/pull_request_number: '{{pull_request_number}}'
+    build.appstudio.redhat.com/target_branch: '{{target_branch}}'
+    pipelinesascode.tekton.dev/max-keep-runs: "3"
+    pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch
+      == "main"
```
amisstea, line 11:
```suggestion
    pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch == "main"
```

.tekton/o11y-push.yaml
hmariset: Updated!

hmariset:
@amisstea Updated the code based on the suggestions. Please review again.

.tekton/o11y-push.yaml
scoheb: why are we not using the release pipeline that includes these types of tasks? other RHTAP components follow this pattern

amisstea: @scoheb good point. We'll switch to using a release pipeline as part of [RHTAPWATCH-704](https://issues.redhat.com/browse/RHTAPWATCH-704).

Since this task will be removed from the push pipeline soon it's not worth worrying about the execution order. Resolving this conversation.

amisstea (approved):

yftacherzog (approved):
lgtm

PR #152, opened 2024-01-04
chore(RHTAPREL-304): add alerting for release service SLOs

theflockers:
add SLO alert rules for the Release Service

- ReleaseServiceProcessingDuration:
  90% of the releases should be processed under 1 hour
- ReleaseServiceStartProcessingSeconds:
  90% of the releases should start processing under 10 seconds
- ReleaseServiceValidationSeconds:
  90% of the releases should be validated under 5 seconds
  
### DIFF
--- a/rhobs/alerting/data_plane/prometheus.release_service_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.release_service_alerts.yaml
@@ -1,0 +1,77 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: rhtap-release-service-alerting
+  labels:
+    tenant: rhtap
+spec:
+  groups:
+  - name: release_service_alerts
+    interval: 1m
+
+    rules:
+    # record rules for `ReleaseServiceProcessingDuration`
+    - record: rate_of_release_processing_duration_seconds_shorter_than_1h_per_10m
+      expr: sum by (job) (rate(release_processing_duration_seconds_bucket{le="3600"}[10m]))
+
+    - record: rate_of_release_processing_duration_seconds_counter_per_10m
+      expr: sum by (job) (rate(release_processing_duration_seconds_count[10m]) > 0)
+
+    # record rules for `ReleaseServicePreProcessingDurationSeconds`
+    - record: rate_of_release_pre_processing_duration_seconds_shorter_than_10s_per_10m
+      expr: sum by (job) (rate(release_pre_processing_duration_seconds_bucket{le="10"}[10m]))
+
+    - record: rate_of_release_pre_processing_duration_seconds_counter_per_10m
+      expr: sum by (job) (rate(release_pre_processing_duration_seconds_count[10m]) > 0)
+
+    # record rules for `ReleaseServiceValidationDurationSeconds`
+    - record: rate_of_release_validation_duration_seconds_shorter_than_5s_per_10m
+      expr: sum by (job) (rate(release_validation_duration_seconds_bucket{le="5"}[10m]))
+
+    - record: rate_of_release_validation_duration_seconds_counter_per_10m
+      expr: sum by (job) (rate(release_validation_duration_seconds_count[10m]) > 0)
+
+    - alert: ReleaseServiceProcessingDuration
+      expr: |
+        (
+          rate_of_release_processing_duration_seconds_shorter_than_1h_per_10m / rate_of_release_processing_duration_seconds_counter_per_10m
+        ) < 0.90
+      for: 15m
+      labels:
+        severity: critical
+        slo: "true"
+      annotations:
+        summary: >-
+          90% of Releases must be processed under one hour
+        description: >-
+          Release service is failing to successfully process within the period of one hour for 90% of releases
+
+    - alert: ReleaseServicePreProcessingDurationSeconds
+      expr: |
+        (
+          rate_of_release_pre_processing_duration_seconds_shorter_than_10s_per_10m / rate_of_release_pre_processing_duration_seconds_counter_per_10m
+        ) < 0.90
+      for: 15m
+      labels:
+        severity: critical
+        slo: "true"
+      annotations:
+        summary: >-
+          90% of Releases must start processing under 10 seconds
+        description: >-
+          Release service is failing to start processing under 10 seconds for 90% of releases
+
+    - alert: ReleaseServiceValidationDurationSeconds
+      expr: |
+        (
+          rate_of_release_validation_duration_seconds_shorter_than_5s_per_10m / rate_of_release_validation_duration_seconds_counter_per_10m
+        ) < 0.90
+      for: 15m
+      labels:
+        severity: critical
+        slo: "true"
+      annotations:
+        summary: >-
+          90% of Releases must be validated under 5 seconds
+        description: >-
+          Release service is failing to run the validations under 5 seconds for 90% of releases
--- a/test/promql/tests/data_plane/release_service_test.yaml
+++ b/test/promql/tests/data_plane/release_service_test.yaml
@@ -1,0 +1,77 @@
+rule_files:
+  - prometheus.release_service_alerts.yaml
+
+evaluation_interval: 1m
+
+tests:
+  - interval: 1m
+    input_series:
+      - series: 'release_processing_duration_seconds_bucket{le="3600", job="release", source_cluster="cluster01"}'
+        values: '1+8x59'
+      - series: 'release_processing_duration_seconds_count{job="release", source_cluster="cluster01"}'
+        values: '1+9x59'
+      - series: 'release_processing_duration_seconds_bucket{le="3600", job="release", source_cluster="cluster02"}'
+        values: '1+8x59'
+      - series: 'release_processing_duration_seconds_count{job="release", source_cluster="cluster02"}'
+        values: '1+9x59'
+    alert_rule_test:
+      - eval_time: 1h
+        alertname: ReleaseServiceProcessingDuration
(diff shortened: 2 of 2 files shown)


theflockers:
Depends on https://github.com/redhat-appstudio/release-service/pull/331

yftacherzog:
@theflockers can you rebase the PR? There were some changes in the tests we're running and until the code is rebased, it's expecting status from tests that do not exist anymore.

test/promql/tests/data_plane/release_service_test.yaml
```diff
@@ -0,0 +1,65 @@
+rule_files:
+  - prometheus.release_service_alerts.yaml
+
+evaluation_interval: 1m
+
+tests:
```
yftacherzog, line 6: All recording rules are using `sum`, but no test case tests that.

rhobs/alerting/data_plane/prometheus.release_service_alerts.yaml
```diff
@@ -0,0 +1,77 @@
...
+  - name: release_service_alerts
+    interval: 1m
+
+    rules:
+    # record rules for `ReleaseServiceProcessingDuration`
+    - record: rate_of_release_processing_duration_seconds_longer_than_1h_per_10m
+      expr: sum by (job) (rate(release_processing_duration_seconds_bucket{le="3600"}[10m]))
+
+    - record: rate_of_release_processing_duration_seconds_counter_per_10m
+      expr: sum by (job) (rate(release_processing_duration_seconds_count[10m]) > 0)
```
yftacherzog, line 18: What's the reason for adding the `> 10` on those rules?

rhobs/alerting/data_plane/prometheus.release_service_alerts.yaml
```diff
@@ -0,0 +1,77 @@
...
+    tenant: rhtap
+spec:
+  groups:
+  - name: release_service_alerts
+    interval: 1m
+
+    rules:
+    # record rules for `ReleaseServiceProcessingDuration`
+    - record: rate_of_release_processing_duration_seconds_longer_than_1h_per_10m
+      expr: sum by (job) (rate(release_processing_duration_seconds_bucket{le="3600"}[10m]))
```
yftacherzog, line 15: The way it is right now, we're summing across clusters. Is that intentional? Should we consider the health of the service per-cluster or for all clusters together?

yftacherzog: We should also have that in mind on the alert rules summary/description. If it should be per-cluster, then we need to tell there which cluster the alert is for.

rhobs/alerting/data_plane/prometheus.release_service_alerts.yaml
```diff
@@ -0,0 +1,77 @@
...
+      expr: sum by (job) (rate(release_validation_duration_seconds_bucket{le="5"}[10m]))
+
+    - record: rate_of_release_validation_duration_seconds_counter_per_10m
+      expr: sum by (job) (rate(release_validation_duration_seconds_count[10m]) > 0)
+
+    - alert: ReleaseServiceProcessingDuration
+      expr: |
+        (
+          rate_of_release_processing_duration_seconds_longer_than_1h_per_10m / rate_of_release_processing_duration_seconds_counter_per_10m
+        ) < 0.90
```
yftacherzog, line 38: Is this the correct operator? We want to alert when this ratio is larger than 0.9. No?

rhobs/alerting/data_plane/prometheus.release_service_alerts.yaml
yftacherzog: Do we want the `release_processing_duration_seconds_bucket{le="3600"}` here or its compliment?
i.e. inf minus 1h.

rhobs/alerting/data_plane/prometheus.release_service_alerts.yaml
theflockers: In this case it is the other way around. It should alert if the majority of the cases goes above the threshold. 
Correct me if my the logic does not represent the calculation:

We have 100 releases and the majority should fall in the buckets smaller than 3600. If get reach 0.90 of the requests falling in the bucket 3600, I want to alert.

test/promql/tests/data_plane/release_service_test.yaml
theflockers: hum. I understood that test case would only need the raw data and the positive (or negative) case and the sum would occur only in the alert rule itself. Why a sum test case would be required? Could you clarify?

rhobs/alerting/data_plane/prometheus.release_service_alerts.yaml
theflockers: I am not sure if this is a feature or a bug but if the `>0` is not added, in case we have no occurrences in the desired bucket the function returns no data (maybe due to the division by 0). By adding it only calculates if there is data available, hence when there is no data, the value is 1.00.

rhobs/alerting/data_plane/prometheus.release_service_alerts.yaml
theflockers: > The way it is right now, we're summing across clusters. Is that intentional? Should we consider the health of the service per-cluster or for all clusters together?

Not intentional. I just have no idea how these metrics are aggregated. In my mind I was adding alerts for "prod" regardless on where it runs.

> Do we want the `release_processing_duration_seconds_bucket{le="3600"}` here or its compliment? i.e. inf minus 1h.
The understanding I had from the prometheus doc was that if a release takes >=3600 secs it will be added here and `+Inf`. So AFAIU I would just need to take in consideration the data within the bucket that interests me. Is that correct?

theflockers:
> @theflockers can you rebase the PR? There were some changes in the tests we're running and until the code is rebased, it's expecting status from tests that do not exist anymore.

sure. On it.

rhobs/alerting/data_plane/prometheus.release_service_alerts.yaml
yftacherzog: Maybe the recording rule should be named `rate_of_release_processing_duration_seconds_shorter_than_1h_per_10m` instead of `...longer_than...`.

rhobs/alerting/data_plane/prometheus.release_service_alerts.yaml
yftacherzog: > Not intentional. I just have no idea how these metrics are aggregated. In my mind I was adding alerts for "prod" regardless on where it runs.

I suggest adding a test case that includes time series from multiple clusters and verify that things are aggregated correctly. Search for `source_cluster` in some of the other test files. You should be able to use something similar.

rhobs/alerting/data_plane/prometheus.release_service_alerts.yaml
yftacherzog: alright then.

rhobs/alerting/data_plane/prometheus.release_service_alerts.yaml
theflockers: Reading now, my last phrase was unclear actually. Is not  "reach" 0.90, but "fall below" 0.90.
The healthy state is between `0.90` and `1.0`. But I see -  although I am searching for pipelineruns in the "+3600" bucket, the result regards to the ones below it. I will change.

rhobs/alerting/data_plane/prometheus.release_service_alerts.yaml
theflockers: I agree. I will add to the cases.

test/promql/tests/data_plane/release_service_test.yaml
theflockers: I see.

yftacherzog (commented):
Overall, this looks good. Tag me when you rebased and done with renaming.

rhobs/alerting/data_plane/prometheus.release_service_alerts.yaml
theflockers: @yftacherzog I added the source_cluster to the existent cases instead of adding more tests. Is that ok for you?

yftacherzog: Are you ok with the current behavior? I suspect that currently you're:
1. summing across clusters, which means it will only trigger when the aggregation across all cluster will apply to the threshold.
2. you're summary/description does not include the cluster label so even if you did not aggregate across multiple clusters, you'd have hard time knowing which cluster the alert is about.

theflockers: Correct me again if I am wrong @yftacherzog: 

I understand that SLO is not bounded to cluster but the service overall, so I think the cluster from where it comes (for SLO only) is not relevant. We might (and will soon) have more specific alerts where this information will be  relevant so we can take a direct action to it.

I added the `source_cluster` only to justify the `sum` in the query. Not a relevant information at this moment as said above. WDYT?

yftacherzog: I'm just highlighting what I see in the implementation. I leave it for you to decide as you have the understanding of the SLO and the requirements.

yftacherzog: Can you rebase so that we don't wait on that check that doesn't exist anymore?

theflockers: done

PR #161, opened 2024-01-09
chore(RHTAPWATCH-530): namespace annotation on pod alerts

Omeramsc:
Adds a "namespace" annotation to all `pod alerts` alerts,
and to `rhtap-gitops-argocd-app-reconcile-alert` alert.

This is used to test a future feature that will tag the team
relevant to the alert where the alert is recieved.
Alert recievers doesn't have access to the prometheus labels,
therefor the label has to be added to the alert itself, which
the reciever can access.

Although the "namespace" is available on some alerts as a default
label, adding it as an annotation for uniformity and readability sake.

### DIFF
--- a/rhobs/alerting/control_plane/prometheus.argocd_alerts.yaml
+++ b/rhobs/alerting/control_plane/prometheus.argocd_alerts.yaml
@@ -24,6 +24,7 @@ spec:
           Application: {{ $labels.name }}
           Cluster: {{ $labels.dest_server }}
           Health status Degraded.
+        alert_route_namespace: '{{ $labels.dest_namespace }}'
         runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/o11y/alert-rule-degradedArgocdApp.md
     - alert: ProgressingArgocdApp
       expr: |
@@ -44,4 +45,5 @@ spec:
           Application: {{ $labels.name }}
           Cluster: {{ $labels.dest_server }}
           App progressing for too long.
+        alert_route_namespace: '{{ $labels.dest_namespace }}'
         runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/o11y/alert-rule-ProgressingArgocdApp.md
--- a/rhobs/alerting/data_plane/prometheus.gitops_argocd_app_reconcile_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.gitops_argocd_app_reconcile_alerts.yaml
@@ -17,6 +17,7 @@ spec:
           Argo CD in namespace {{ $labels.namespace }}
           has failed to reconcile at least 99% of applications
           within 4 seconds over the past 30 minutes
+        alert_route_namespace: gitops-service-argocd
         runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/gitops/argocd-app-reconcile.md
       expr: |
         increase(argocd_app_reconcile_bucket{namespace="gitops-service-argocd", le="4"}[30m])
--- a/rhobs/alerting/data_plane/prometheus.pod_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.pod_alerts.yaml
@@ -19,6 +19,7 @@ spec:
         description: >-
           Pod {{ $labels.pod }} for namespace {{ $labels.namespace }} on cluster
           {{ $labels.source_cluster }} is unscheduled for more than 30 minutes.
+        alert_route_namespace: '{{ $labels.namespace }}'
         runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/o11y/alert-rule-unschedualablePods.md
     - alert: CrashLoopBackOff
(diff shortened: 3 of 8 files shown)


rhobs/alerting/data_plane/prometheus.pod_alerts.yaml
```diff
@@ -19,6 +19,7 @@ spec:
         description: >-
           Pod {{ $labels.pod }} for namespace {{ $labels.namespace }} on cluster
           {{ $labels.source_cluster }} is unscheduled for more than 30 minutes.
+        dest_namespace: {{ $labels.dest_namespace }}
```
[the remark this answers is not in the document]
Omeramsc: We can use `namespace` for the annotation name, I assumed we'll be using `dest_namespace`, therefore I preferred it to be less confusing and more straightforward. as this doesn't seem to be the case, `namespace` is good enough.

rhobs/alerting/data_plane/prometheus.gitops_argocd_app_reconcile_alerts.yaml
```diff
@@ -17,6 +17,7 @@ spec:
           Argo CD in namespace {{ $labels.namespace }}
           has failed to reconcile at least 99% of applications
           within 4 seconds over the past 30 minutes
+        namespace: '{{ if .Labels.dest_namespace }}{{ .Labels.dest_namespace }}{{ else if .Labels.namespace }}{{ .Labels.namespace }}{{ else }}""{{ end }}'
```
yftacherzog, line 20: Does  `dest_namespace` only exists some of the times?
Also, the expression depends on `namespace=gitops-service-argocd` so that label should always exist if the alert is firing.

I.e. I don't think we need the last else, and maybe neither the first.

If we keep the `else`, we need test cases that cover them.

Omeramsc: misunderstanding on my part, I don't know of situations when the `dest_namespace` are absent. refined, and added the required label to the time series.

rhobs/alerting/data_plane/prometheus.gitops_argocd_app_reconcile_alerts.yaml
amisstea, line 1: This is one of the gitops service SLO alerting rules for the data plane (`slo: "true"` label applied). We already know who should (eventually) be tagged for such alerts: the gitops service team. `dest_namespace` in this case is likely a tenant namespace for deploying ephemeral envs so it won't be useful for mapping namespaces to teams.

Instead, we should add the `namespace` annotation to the control plane [ArgoCD alerting rules](https://github.com/redhat-appstudio/o11y/blob/main/rhobs/alerting/control_plane/prometheus.argocd_alerts.yaml). In those the `dest_namespace` label should indicate where the various konflux services are being deployed.

rhobs/alerting/data_plane/prometheus.gitops_argocd_app_reconcile_alerts.yaml
```diff
@@ -17,6 +17,7 @@ spec:
           Argo CD in namespace {{ $labels.namespace }}
           has failed to reconcile at least 99% of applications
           within 4 seconds over the past 30 minutes
+        namespace: '{{ .Labels.dest_namespace }}'
```
amisstea, line 20: I haven't tested this but shouldn't this use `$labels`?
```suggestion
        namespace: '{{ $labels.dest_namespace }}'
```

amisstea (dismissed):

rhobs/alerting/data_plane/prometheus.gitops_argocd_app_reconcile_alerts.yaml
Omeramsc: fixed

rhobs/alerting/data_plane/prometheus.gitops_argocd_app_reconcile_alerts.yaml
Omeramsc: I added the annotation logic to the control plane
I kept a hardcoded namespace on the SLO alert rule as that is another case we should take care of anyway

rhobs/alerting/control_plane/prometheus.argocd_alerts.yaml
```diff
@@ -24,6 +24,7 @@ spec:
           Application: {{ $labels.name }}
           Cluster: {{ $labels.dest_server }}
           Health status Degraded.
+        namespace: '{{ $labels.dest_namespace }}'
```
yftacherzog, line 27: Maybe we can name this annotation differently somehow? Currently it will mean that there's a namespace label and a namespace annotation with different values, which might be confusing.

Maybe can give it a name that will suggest what it's for. e.g. `alert_routing_namespace` or something along those lines).

Omeramsc: WDYT about `alert_namespace_tag` or `alert_namespace_route`?

yftacherzog: I think it should end with `namespace`, and whatever comes before it should describe the namespace. Can be `alert_route_namespace`.

Omeramsc: changed

yftacherzog (approved):
lgtm

PR #168, opened 2024-01-18
chore(RHTAPWATCH-675): Update annotations for control_plane alerts

avi-biton:
Adding a `summary` annotation and rename the `message` annotation to `description` to be alligned with all the alerts in `o11y` and to make it easier to extract these values when we route the alerts.

### DIFF
--- a/rhobs/alerting/control_plane/prometheus.argocd_alerts.yaml
+++ b/rhobs/alerting/control_plane/prometheus.argocd_alerts.yaml
@@ -19,7 +19,8 @@ spec:
         severity: warning
         team: rhtap
       annotations:
-        message: |
+        summary: Argocd App '{{ $labels.name }}' is degraded
+        description: |
           Environment: {{ $labels.source_environment }}
           Application: {{ $labels.name }}
           Cluster: {{ $labels.dest_server }}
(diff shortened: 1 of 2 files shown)


Omeramsc (approved):
LGTM

rhobs/alerting/control_plane/prometheus.argocd_alerts.yaml
```diff
@@ -19,7 +19,8 @@ spec:
         severity: warning
         team: rhtap
       annotations:
-        message: |
+        summary: Argocd App '{{ $labels.name }}' is degraded
+        description: |
```
yftacherzog, line 23: In case we'd like to preserve the option to do things as we did before, is it possible to have something like:
```
        summary: Argocd App '{{ $labels.name }}' is degraded
        message: |
          Environment: {{ $labels.source_environment }}
          Application: {{ $labels.name }}
          Cluster: {{ $labels.dest_server }}
        description: {{ $annotations.message }}
```
i.e. copy the content of message to description.

avi-biton: It might be possible, but what is the benefit?
We are not using the `message` anywhere, we might as well switch it to `description` to be aligned

yftacherzog: It won't be the first time we move back and forth. Not a must, but figuring out what each system routing template is expecting is a headache. I think that if we can keep both options possible without code duplication, then maybe we should do it.

avi-biton: I tried adding it as you described, it seems that using `$annotations` within the `annotation` section is not working.
I think we will leave it as it is for now.

PR #169, opened 2024-01-18
test(RHTAPWATCH-637): always-firing alert to test tagging logic

Omeramsc:
Adds an alert that always firing in order to test the slack tagging logic sooner than later.

Will probably change the annotation more later to
test different cases.


yftacherzog (approved):
lgtm

PR #170, opened 2024-01-18
test(RHTAPWATCH-637): always-firing directory change fix

Omeramsc:
moving the always-firing alert to the correct directory


avi-biton (approved):
lgtm

PR #174, opened 2024-01-19
feat(RHTAPWATCH-570): Update exporters code with actual metrics

hmariset:
Update the dsexporter code and related unit tests with actual metrics.


exporters/dsexporter/dsexporter.go
```diff
@@ -3,37 +3,69 @@ package main
...
+	service := req.URL.Query().Get("service")
+	check := req.URL.Query().Get("check")
+	availability := checkIfServiceUp(service, check)
+
+	exporter.appstudioProbeSuccess.WithLabelValues("service", "check").Set(availability)
+	fmt.Fprintf(response, "Service: %s, Check Name: %s, Availability: %f", service, check, availability)
+}
+
+func checkIfServiceUp(service, check string) float64 {
+	serviceUrl := "http://" + service + ".com"
```
amisstea, line 51: Building the url like this makes a lot of assumptions. It'd be simpler to pass the url in directly.

Also, please use `https` by default.

exporters/dsexporter/dsexporter.go
```diff
@@ -3,37 +3,69 @@ package main
...
 func NewCustomCollector() *CustomCollector {
 	return &CustomCollector{
-		requestCounter: prometheus.NewCounter(prometheus.CounterOpts{
-			Name: "request_count",
-			Help: "Number of requests handled by the handler",
-		}),
+		appstudioProbeSuccess: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "appstudio_probe_success",
+			Help: "Availability of appstudio service/component",
+		},
```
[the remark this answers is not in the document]
hmariset: Updated

exporters/dsexporter/dsexporter_test.go
```diff
@@ -16,31 +18,48 @@ func TestCustomCollector(t *testing.T) {
...
 			break
 		}
 	}
 
 	// Check whether the exported metric value is now 1 after incrementing.
-	assert.Equal(t, float64(1), RequestCountValue)
+	assert.Equal(t, float64(1), konfluxProbeSuccessValue)
+}
+
+func TestcheckIfServiceUp(t *testing.T) {
```
yftacherzog, line 50: Need to also test the scenario in which the service is not up.

yftacherzog (changes requested):
Go checks are still failing.

exporters/dsexporter/dsexporter_test.go
hmariset: Updated

exporters/dsexporter/dsexporter_test.go
```diff
@@ -1,46 +1,83 @@
...
-	var RequestCountValue float64
-	for _, mf := range metricFamilies {
-		if mf.GetName() == "request_count" {
-			RequestCountValue = mf.GetMetric()[0].GetCounter().GetValue()
-			break
-		}
+func TestAvailabilityHandler(t *testing.T) {
+	mockExporter := NewCustomCollector()
+
+	req, err := http.NewRequest("GET", "/?service=test.com&check=test", nil)
```
amisstea, line 16: According to the metric [spec](https://docs.google.com/document/d/193FO8DYP8zCzzued53DiIT84xmyTm51v_OqiWzQi8Bg), the `service` label should be an identifier for the service under evaluation, not the lower level check of its dependency (`test.com` as provided here). It should be a constant value shared across the entire exporter (e.g. `dummy-service`) and not provided dynamically.

Also, wdyt about adding more checks (like `github` which could query `https://api.github.com/`) to make this example more realistic? `test` isn't an obvious enough example on its own IMO.

exporters/dsexporter/dsexporter.go
```diff
@@ -3,37 +3,70 @@ package main
...
+func availabilityHandler(response http.ResponseWriter, req *http.Request, exporter *CustomCollector) {
+	service := req.URL.Query().Get("service")
+	check := req.URL.Query().Get("check")
+	availability := checkIfServiceUp(service)
+
+	exporter.konfluxProbeSuccess.WithLabelValues(service, check).Set(availability)
+	fmt.Fprintf(response, "Service: %s, Check Name: %s, Availability: %f", service, check, availability)
+}
+
+func checkIfServiceUp(service string) float64 {
```
amisstea, line 50:
```suggestion
func checkIfDependencyUp(url string) float64 {
```

Using service in the name or args might cause confusion with the metric's `service` label.

We should also pass in the full URL and not make any assumptions about what to query. Allowing the caller to pass an arbitrary host string opens this up to potential abuse. You can maintain a mapping somewhere for `check` to URL.

amisstea (changes requested):

exporters/dsexporter/dsexporter_test.go
yftacherzog: @hmariset I think that having a clear idea (and having documented on the PR description and on the ticket) regarding what sort of metrics we want to include here will be super important for being able to deliver on it.

My take is that when we named it dsexporter, we were planning to use it to export the availability of the grafana data source. We also have a working example for that code. I think we should see that we're able to achieve that.

exporters/dsexporter/dsexporter.go
```diff
@@ -1,39 +1,115 @@
...
+		if strings.Contains(datasource, dsToCheck) {
+			fmt.Println("Datasource", datasource, "exists")
+			return 1
+		}
+	}
+	fmt.Println("Datasource", dsToCheck, "does not exist")
+	return 0
+}
+
+func NewKubeClient() *kubernetes.Clientset {
```
yftacherzog, line 101: Have a look at [GetConfigOrDie](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client/config#GetConfigOrDie). I think it does what you're doing here.

exporters/dsexporter/dsexporter.go
```diff
@@ -1,39 +1,115 @@
...
+	// creates the in-cluster config
+	config, err := rest.InClusterConfig()
```
yftacherzog, line 103: Have a look at [GetConfigOrDie](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client/config#GetConfigOrDie). I think it does what you're doing here.

exporters/dsexporter/dsexporter.go
```diff
@@ -1,39 +1,115 @@
...
+		Get().
+		AbsPath("/apis/grafana.integreatly.org/v1beta1").
+		Namespace("appstudio-grafana").
+		Resource("grafanas").
+		Name("grafana-oauth").
+		DoRaw(context.TODO())
+	var grafanaResoruce map[string]interface{}
+	err = json.Unmarshal(data, &grafanaResoruce)
+	if err != nil {
+		fmt.Printf("Error getting resource: %v\n", err)
```
yftacherzog, line 68: I created a ticket to replace printings with proper logger calls.

exporters/dsexporter/dsexporter.go
```diff
@@ -1,39 +1,115 @@
...
+}
+
+func NewKubeClient() *kubernetes.Clientset {
+	// creates the in-cluster config
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		panic(err.Error())
+	}
+	// creates the clientset
+	clientset, err := kubernetes.NewForConfig(config)
```
yftacherzog, line 108: There's also [NewForConfigOrDie](https://pkg.go.dev/k8s.io/client-go/kubernetes#NewForConfigOrDie).
If you combine this one and `GetConfigOrDie`, I think you can replace this function with a one-liner which also does not require error handling.

exporters/dsexporter/dsexporter.go
```diff
@@ -1,39 +1,115 @@
...
+	datasourcesIfc := grafanaResource["status"].(map[string]any)["datasources"].([]interface{})
+	datasources := make([]string, len(datasourcesIfc))
+	for i, v := range datasourcesIfc {
+		datasources[i] = v.(string)
+	}
+	return datasources
+}
+
+// check if datasource exists, return 1 if yes, 0 if not
+func IsDataSourceExist(datasources []string, dsToCheck string) float64 {
```
yftacherzog, line 90: `IsSomething` suggests a boolean return value. If we name it this way, we should return a boolean and do the translation later.

If we want a function that returns the availability then we should name it accordingly.

exporters/dsexporter/dsexporter.go
```diff
@@ -1,39 +1,115 @@
...
 
 // Creating a new instance of CustomCollector.
 func NewCustomCollector() *CustomCollector {
 	return &CustomCollector{
-		requestCounter: prometheus.NewCounter(prometheus.CounterOpts{
-			Name: "request_count",
-			Help: "Number of requests handled by the handler",
-		}),
+		konfluxProbeSuccess: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "konflux_probe_success",
```
yftacherzog, line 29: We dropped the use of the word `probe` as far us I understand. I think we should name it differently.

exporters/dsexporter/dsexporter.go
```diff
@@ -1,39 +1,115 @@
...
+	var availability float64
+
+	availability = IsDataSourceExist(GetDataSources(GetGrafanaResource()), check)
+	e.konfluxProbeSuccess.WithLabelValues(service, check).Set(availability)
+	e.konfluxProbeSuccess.Collect(ch)
+}
+
+// get the grafna resource as a map
+func GetGrafanaResource() map[string]interface{} {
+	clientset := NewKubeClient()
```
yftacherzog, line 56: I think passing the client as an input will make it easier to test.

exporters/dsexporter/dsexporter.go
```diff
@@ -1,39 +1,115 @@
...
+	"k8s.io/client-go/rest"
 )
 
+const service = "https://console-openshift-console.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com/k8s/ns/appstudio-grafana/services"
+const check = "prometheus-appstudio-ds"
+var kubeconfig *string
+
 type CustomCollector struct {
-	requestCounter prometheus.Counter
+	konfluxProbeSuccess *prometheus.GaugeVec
```
yftacherzog, line 22: What's the incentive for using a vector for gauges here rather than a gauge?

exporters/dsexporter/dsexporter.go
```diff
@@ -1,39 +1,115 @@
...
+	"os"
+	"strings"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 )
 
+const service = "https://console-openshift-console.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com/k8s/ns/appstudio-grafana/services"
```
yftacherzog, line 17: This is going to be different for each cluster. No?

hmariset: Yes, it's different for each cluster

exporters/dsexporter/dsexporter.go
hmariset: Since we have labels associated with the metric I have used gaugeVec

exporters/dsexporter/dsexporter.go
```diff
@@ -1,39 +1,109 @@
...
 
 // Creating a new instance of CustomCollector.
 func NewCustomCollector() *CustomCollector {
 	return &CustomCollector{
-		requestCounter: prometheus.NewCounter(prometheus.CounterOpts{
-			Name: "request_count",
-			Help: "Number of requests handled by the handler",
-		}),
+		konfluxDSExporter: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "konflux_dsexporter",
```
yftacherzog, line 31: I don't think it makes sense using the `exporter` keyword in a metric name. @amisstea worked on the ticket for the metrics specs, and looking at [the outcome](https://docs.google.com/document/d/193FO8DYP8zCzzued53DiIT84xmyTm51v_OqiWzQi8Bg/edit?usp=sharing), I see that the example there was using the `probe` keyword.

Alex, do you think we should use the `probe` keyword? Any other suggestions?

exporters/dsexporter/dsexporter.go
```diff
@@ -1,39 +1,109 @@
...
+			Help: "Availability of konflux service/component",
```
avi-biton, line 32: The help is for the specific metric, I think it should be more informative

```suggestion
			Help: "Availability of prometheus-appstudio-ds Grafana datasource",
```

exporters/dsexporter/dsexporter.go
amisstea: We discussed this in a call. The `service` label should be a constant descriptor for the service under evaluation and it should not be cluster scoped. In this case I think we could use `grafana` since we're checking the availability of grafana-related configuration.

exporters/dsexporter/dsexporter.go
amisstea: As discussed in our paired programming call, let's use `konflux_up` for the metric name (inspired by the prometheus `up` metric). I'll update that spec doc.

exporters/dsexporter/dsexporter.go
amisstea: Or if we don't want to expose `appstudio` unecessarily (since it's an old name):

```suggestion
			Help: "Availability of the Konflux default grafana datasource",
```

exporters/dsexporter/dsexporter.go
hmariset: Updated

exporters/dsexporter/dsexporter.go
hmariset: Updated the code

exporters/dsexporter/dsexporter.go
hmariset: Updated

exporters/dsexporter/dsexporter_test.go
```diff
@@ -2,45 +2,136 @@ package main
...
+    defer func() {
+        getInClusterConfig = actualInclusterConfig
+        getNewForConfig = actualNewForConfig
+    }()
 
-	// Simulate collecting metrics and check the exported metric value.
-	metrics := prometheus.DefaultGatherer
-	metricFamilies, err := metrics.Gather()
-	assert.NoError(t, err)
+    client := NewKubeClient()
```
yftacherzog, line 36: Once we drop all the error handling, I don't think we need a function for this, and I don't think it requires any tests. Can be just a one-liner.

```suggestion
    client := kubernetes.NewForConfigOrDie(config.GetConfigOrDie())
```
Also, the config function actually supports both local and in-cluster configs, so the name/comments might be a bit misleading.

exporters/dsexporter/dsexporter_test.go
```diff
@@ -2,45 +2,136 @@ package main
...
+func MockInclusterConfig() (*rest.Config) {
+    return nil
+}
+
+func MockNewForConfig(*rest.Config) (*kubernetes.Clientset) {
+	return nil
+}
+
+func TestNewKubeClient(t *testing.T) {
+	actualInclusterConfig := getInClusterConfig
```
yftacherzog, line 26: Fix indentation

exporters/dsexporter/dsexporter_test.go
```diff
@@ -2,45 +2,136 @@ package main
...
+    getInClusterConfig = MockInclusterConfig
+    actualNewForConfig := getNewForConfig
+    getNewForConfig = MockNewForConfig
+
+    defer func() {
+        getInClusterConfig = actualInclusterConfig
+        getNewForConfig = actualNewForConfig
+    }()
```
yftacherzog, line 34: What's the reason for having this?

exporters/dsexporter/dsexporter.go
```diff
@@ -1,39 +1,108 @@
...
+	for i, v := range datasourcesIfc {
+		datasources[i] = v.(string)
+	}
+	return datasources
+}
+
+// check if datasource exists, return 1 if yes, 0 if not
+func CheckDataSourceExist(datasources []string, dsToCheck string) float64 {
+	for _, datasource := range datasources {
+		if strings.Contains(datasource, dsToCheck) {
```
yftacherzog, line 91: What's the reason for using `Contains` rather than checking if the strings are equal?

yftacherzog: I think we should be able to replace this entire loop a call to `slices.Contains`.
See here: https://stackoverflow.com/a/71181131

exporters/dsexporter/dsexporter.go
```diff
@@ -1,39 +1,108 @@
...
+		os.Exit(1)
+	}
+
+	return grafanaResource
+}
+
+// get datasources from grafana resource
+func GetDataSources(grafanaResource map[string]interface{}) []string {
+	// return empty string slice if datasources are not defined
+	if grafanaResource["status"].(map[string]any)["datasources"] == nil {
```
yftacherzog, line 77: Are we covering the case in which the map does not include `status` at all?

exporters/dsexporter/dsexporter.go
```diff
@@ -1,39 +1,108 @@
...
+		return make([]string, 0)
+	}
+	datasourcesIfc := grafanaResource["status"].(map[string]any)["datasources"].([]interface{})
```
yftacherzog, line 80: What is this line doing? How does this work?

exporters/dsexporter/dsexporter.go
```diff
@@ -1,39 +1,108 @@
...
+	datasources := make([]string, len(datasourcesIfc))
+	for i, v := range datasourcesIfc {
+		datasources[i] = v.(string)
```
yftacherzog, line 83: What is this line doing? How does this work?

exporters/dsexporter/dsexporter_test.go
yftacherzog: Maybe this line is actually with the correct indentation and the ones below are under-indented. Need to keep it consistent across the code.

exporters/dsexporter/dsexporter.go
hmariset: Here the datasource we are getting is `appstudio-grafana/prometheus-appstudio-ds/b80cfcd5-3010-4e6a-be44-6843e0fef2d8` so when we are checking for `prometheus-appstudio-ds` `Contains` returns true. Also, `slices.Contains` is checking for exact match in the string so it's not working.

exporters/dsexporter/dsexporter_test.go
hmariset: it ensures original functions are restored and acts like a cleanup

exporters/dsexporter/dsexporter.go
hmariset: it's a string type assertion. If value is not string it throws a panic error.

exporters/dsexporter/dsexporter.go
hmariset: it's type assertions. If the type doesn't match with the type mentioned over there it would throw a panic error.

exporters/dsexporter/dsexporter.go
yftacherzog: > Here the datasource we are getting is appstudio-grafana/prometheus-appstudio-ds/b80cfcd5-3010-4e6a-be44-6843e0fef2d8

We should let the test case reflect that so the structure of the inputs is clear.

> Also, slices.Contains is checking for exact match in the string so it's not working.

If we cannot use exact match, then we should narrow down the chances of false positives. Either use some sort of regex or use the expected format of the datasource that is returning to make the chances of false match slimmer.

exporters/dsexporter/dsexporter.go
yftacherzog: I didn't know that. Thanks!

exporters/dsexporter/dsexporter.go
yftacherzog: Thanks

exporters/dsexporter/dsexporter_test.go
yftacherzog: I think now that the problem is that you have a combination of tabs and spaces. You can configure your editor to automatically convert one to the other.
I think code bases that we're involved with are using spaces more often than tabs, so better use spaces all around and set the editor to convert tabs to spaces as you type them.

exporters/dsexporter/dsexporter_test.go
yftacherzog: I think it makes sense for this function to take the client as input rather than creating it and then worrying about keeping the integrity of this global variable at the scope of this function.

exporters/dsexporter/dsexporter_test.go
```diff
@@ -2,45 +2,136 @@ package main
...
-			break
-		}
+	expectedResult = []string{}
+	result = GetDataSources(grafanaRes)
+	if !reflect.DeepEqual(result, expectedResult) {
+		t.Errorf("Test-2 failed, Expected %v, but got %v", expectedResult, result)
+	}
+}
+
+func TestMain(t *testing.T) {
```
yftacherzog, line 108: What are we testing here? Is our code being called?

exporters/dsexporter/dsexporter.go
hmariset: Updated

exporters/dsexporter/dsexporter_test.go
hmariset: here the server URL is some mock URL but we have exporter defined here same as our actual code and checks whether it's setting the labels and availability value correctly.

exporters/dsexporter/dsexporter.go
hmariset: Updated test case and using regex for matching.

exporters/dsexporter/dsexporter_test.go
hmariset: removed this test case since modified the code

exporters/dsexporter/dsexporter_test.go
hmariset: Updated.

exporters/dsexporter/dsexporter_test.go
yftacherzog: I think the tests should run our actual code. Testing a copy of the code inside a test does not ensure we're testing the same thing and it also does not protect us from regressions (which is the main purpose of having unit tests in the first place).

Suppose that we make a change that breaks the, the test will still pass as it runs the old code.

All in all, we need to write the code in such a way that will allow us to cover as much of it with unit tests as possible. main functions are sometime trickier to test, so we need to keep them as simple as possible.

exporters/dsexporter/dsexporter.go
```diff
@@ -1,39 +1,106 @@
...
+	e.konfluxUp.Describe(ch)
 }
 
 // Collect method sends the current values of the metrics to Prometheus.
 // After Prometheus understands what metrics are available (using the `Describe` method),
 // it then calls the `Collect` method to actually get the values of those metrics.
 func (e *CustomCollector) Collect(ch chan<- prometheus.Metric) {
-	e.requestCounter.Collect(ch)
+	var availability float64
+	dsToCheck := ".*/" + check + "/.*"
```
yftacherzog, line 47: Checking for regex `.*/SOMETHING/.*` is the same as checking if the string contains `/SOMETHING/`. If this is what we're checking then I'd prefer using contains rather than introducing a regex. We should only introduce regex when it provides something that we cannot do without it (as they are hard to maintain).

avi-biton: In this case the regex does not add any value and the `contain` is easier to understand

exporters/dsexporter/dsexporter_test.go
```diff
@@ -2,45 +2,113 @@ package main
...
+	exporter := prometheus.NewGaugeVec(
```
avi-biton, line 86: Can't we use the `NewCustomCollector` instead of copy it's content?

exporters/dsexporter/dsexporter.go
hmariset: Updated to `Contains` again

exporters/dsexporter/dsexporter.go
hmariset: Updated back to `strings.Contains`  again

exporters/dsexporter/dsexporter_test.go
hmariset: Updated to use `NewCustomCollector`

exporters/dsexporter/dsexporter_test.go
hmariset: Updated the test case to call `NewCustomCollector` and run our code.

hmariset:
@yftacherzog @avi-biton @amisstea updated the code. Please review again.

yftacherzog:
> @yftacherzog @avi-biton @amisstea updated the code. Please review again.

Did you confirm that this code can be deployed to the cluster at its current state, and Prometheus is scraping it correctly? I.e. you are able to query for the new metric and if you kill the grafana operator/service the metric reflects that.

hmariset:
> > @yftacherzog @avi-biton @amisstea updated the code. Please review again.
> 
> Did you confirm that this code can be deployed to the cluster at its current state, and Prometheus is scraping it correctly? I.e. you are able to query for the new metric and if you kill the grafana operator/service the metric reflects that.

@yftacherzog Yes, Prometheus is scraping correctly. I'm able to query the new metric and I also modified the metric to reflect 0 when grafana operator is killed.

exporters/dsexporter/dsexporter.go
```diff
@@ -1,39 +1,108 @@
...
+		AbsPath("/apis/grafana.integreatly.org/v1beta1").
+		Namespace("appstudio-grafana").
+		Resource("grafanas").
+		Name("grafana-oauth").
+		DoRaw(context.TODO())
+	var grafanaResource map[string]interface{}
+	err = json.Unmarshal(data, &grafanaResource)
+	if err != nil {
+		fmt.Printf("Error getting resource: %v\n", err)
+		os.Exit(1)
```
hmariset, line 72: @yftacherzog @avi-biton @amisstea Here as well should I modify to return error message and then return metric value as 0 instead of exiting?

yftacherzog:
> > > @yftacherzog @avi-biton @amisstea updated the code. Please review again.
> > 
> > Did you confirm that this code can be deployed to the cluster at its current state, and Prometheus is scraping it correctly? I.e. you are able to query for the new metric and if you kill the grafana operator/service the metric reflects that.
> 
> @yftacherzog Yes, Prometheus is scraping correctly. I'm able to query the new metric and I also modified the metric to reflect 0 when grafana operator is killed.

I'm asking whether you confirmed it actually shows 0 correctly.

exporters/dsexporter/dsexporter.go
yftacherzog: Yes. I think that's the way to go.

hmariset: Updated it.

hmariset:
> > > > @yftacherzog @avi-biton @amisstea updated the code. Please review again.
> > > 
> > > 
> > > Did you confirm that this code can be deployed to the cluster at its current state, and Prometheus is scraping it correctly? I.e. you are able to query for the new metric and if you kill the grafana operator/service the metric reflects that.
> > 
> > 
> > @yftacherzog Yes, Prometheus is scraping correctly. I'm able to query the new metric and I also modified the metric to reflect 0 when grafana operator is killed.
> 
> I'm asking whether you confirmed it actually shows 0 correctly.

I confirm it shows 0 when I killed the grafana operator/service.

amisstea (approved):
LGTM

yftacherzog (approved):
lgtm

avi-biton (approved):
/lgtm

PR #176, opened 2024-01-22
fix(RHTAPWATCH-739): ignore tekton-ci and build-definitions-e2e

amisstea:
This should reduce exessive alert noise originating from namespaces used to build and test Konflux components.


hmariset (approved):
lgtm

PR #182, opened 2024-01-24
fix(RHTAPWATCH-757): Don't monitor the build-templates-e2e namespace

amisstea:
This corrects a simple mistake from RHTAPWATCH739.


hmariset (approved):
LGTM

PR #198, opened 2024-02-09
fix(STONEINTG-777): extend the window for PipelineRunToSnapshotExceeded

dirgim:
* The alert is currently flaky and flapping because of different services taking longer to finalize the build pipelineRuns
* See also [KFLUXBUGS-24](https://issues.redhat.com/browse/KFLUXBUGS-24)

### DIFF
--- a/rhobs/alerting/data_plane/prometheus.pipeline_to_snapshot_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.pipeline_to_snapshot_alerts.yaml
@@ -13,13 +13,13 @@ spec:
       expr: |
        (
         (
-         increase(integration_svc_response_seconds_bucket{le="+Inf"}[5m])
+         increase(integration_svc_response_seconds_bucket{le="+Inf"}[10m])
          - ignoring(le)
-         increase(integration_svc_response_seconds_bucket{le="30"}[5m])
+         increase(integration_svc_response_seconds_bucket{le="30"}[10m])
         ) / ignoring(le)
-        increase(integration_svc_response_seconds_bucket{le="+Inf"}[5m])
+        increase(integration_svc_response_seconds_bucket{le="+Inf"}[10m])
        ) > 0.10
-      for: 1m
+      for: 10m
       labels:
         severity: critical
         slo: "true"
@@ -28,7 +28,7 @@ spec:
           PipelineRunFinish to SnapshotInProgress time exceeded
         description: >
           Time from pipeline run finished to snapshot marked in progress has been over
-          30s for more than 10% of requests during the last 5 minutes on cluster
-          {{ $labels.source_cluster }}
+          30s for more than 10% of requests during the last 10 minutes and has not improved for 10 minutes
+          on cluster {{ $labels.source_cluster }}
         alert_routing_key: integration-service
         runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/integration-service/pipeline_to_snapshot_exceeded.md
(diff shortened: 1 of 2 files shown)


gabemontero (commented):
if this is a timing / flake sort of thing, I think I would also give a larger buffer to the condition to persist before the alert is fired

i.e. change `for: 1m` on line 22 to say `for: 10m`

gabemontero (approved):

rhobs/alerting/data_plane/prometheus.pipeline_to_snapshot_alerts.yaml
```diff
@@ -28,7 +28,7 @@ spec:
           PipelineRunFinish to SnapshotInProgress time exceeded
         description: >
           Time from pipeline run finished to snapshot marked in progress has been over
-          30s for more than 10% of requests during the last 5 minutes on cluster
+          30s for more than 10% of requests during the last 10 minutes on cluster
```
yftacherzog, line 31: We're looking now at a time window of 10 minutes backward and we require it to fulfill a condition continuously for 10 minutes.
This is different comparing to what we had before in which we looked at a window of 5 minutes and we required the condition to be consistent for 1 minute.

The change in the description might need to reflect that.

test/promql/tests/data_plane/pipeline_to_snapshot_test.yaml
```diff
@@ -19,7 +19,7 @@ tests:
     - series: 'integration_svc_response_seconds_bucket{le="+Inf", namespace="integration-service", source_cluster="cluster02"}'
       values: '0+100x10'  # 100 total occurrences in cluster2
   alert_rule_test:
-    - eval_time: 10m
+    - eval_time: 15m
```
yftacherzog, line 22: I think we should also add a test case in which we test the entire span of the window + the "for".

I.e. a test case in which we evaluate in 21 minutes and we have a bucket which started misbehaving at minute 7, and another that started misbehaving in the same way at minute 0 and we get the alert for the second one and not for the first.

rhobs/alerting/data_plane/prometheus.pipeline_to_snapshot_alerts.yaml
dirgim: Ack, updated the description to account for the condition to be failing continuously for 10 minutes.

test/promql/tests/data_plane/pipeline_to_snapshot_test.yaml
```diff
@@ -31,11 +31,42 @@ tests:
...
               cluster01
             alert_routing_key: integration-service
             runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/integration-service/pipeline_to_snapshot_exceeded.md
 
+  # Scenario where one cluster crosses the 10% threshold after 7 minutes and another does not
+- interval: 1m
+  input_series:
+    # Simulating data from Cluster 1
+    - series: 'integration_svc_response_seconds_bucket{le="30", namespace="integration-service",  source_cluster="cluster01"}'
+      values: '7+10x10'  # 10 requests took less than 30s
```
yftacherzog, line 44:
```suggestion
      values: '7+10x20'
```

We now evaluate at 21 minutes, but we have time series for only 11 minutes. We need to extend the time series.  I see now that other test cases were not updated with longer time series either.

yftacherzog (approved):
lgtm

amisstea (approved):
/lgtm

PR #201, opened 2024-02-14
chore(RHTAPWATCH-823): SLO alerts uses team handles for routing

Omeramsc:
Now that SLO and non-SLO alerts are routed to different channels, there is no need to keep track of the namespace of each team. Instead, alert_routing_key is replaced by the slack's team handle, using the key:value `alert_team_handle`.

Since the team-handle for tagging in slack is using the team ID, a `team` key is also added to make the routing more readable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,21 +63,22 @@ where the team's handle is tagged in the alert message.
 
 #### Usage Guidelines:
 
-Apply the `alert_routing_key` annotation to alerts in order to get notified about them.
+Apply the `alert_team_handle` and `team` annotations to SLO alerts in order to get notified about them.
   
-#### How to apply the `alert_routing_key` Annotation:
+#### How to apply the `alert_team_handle` Annotation:
 
-Apply the `alert_routing_key` key to the annotations section of any alerting rule,
-with one of the team's namespaces as its value, or the team's name.
+Apply the `alert_team_handle` key to the annotations section of any alerting rule,
+with the relevant team's Slack group handle.
+The format of the Slack handle is: `<!subteam^-slack_group_id->` (e.g: `<!subteam^S041261DDEW>`);
+To obtain the Slack group ID, click on the team's group handle, then click the three dots, and select "Copy group ID."
+
+make sure to also add the `team` annotation with the name of the relevant team for readability.
   ```
   annotations:
       summary: "PipelineRunFinish to SnapshotInProgress time exceeded"
-      alert_routing_key: "application-service"
+      alert_team_handle: <!subteam^S04S21ECL8K>
+      team: o11y
   ```
-
-make sure that the team's name is aligned with the one mentioned in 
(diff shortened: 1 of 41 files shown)


README.md
```diff
@@ -63,22 +63,21 @@ where the team's handle is tagged in the alert message.
...
+Apply the `alert_team_handle` and `team` annotations to SLO alerts in order to get notified about them.
   
-#### How to apply the `alert_routing_key` Annotation:
+#### How to apply the `alert_team_handle` Annotation:
 
-Apply the `alert_routing_key` key to the annotations section of any alerting rule,
-with one of the team's namespaces as its value, or the team's name.
+Apply the `alert_team_handle` key to the annotations section of any alerting rule,
+with the relevant team's Slack handle.
+The format of the Slack handle is: `<!subteam^-team_id->` (e.g: `<!subteam^S041261DDEW>`).
```
yftacherzog, line 72: What's `team-id`? Is that the Slack group ID? We should use the same name used in Slack and maybe also explain how to retrieve it.

rhobs/alerting/data_plane/prometheus.latency_image_repository_provision_alerts.yaml
```diff
@@ -6,29 +6,30 @@ metadata:
...
-      annotations:
-        summary: >-
-          Latency of image repository provision time exceeded
-        description: >
-          Time taken to provision image repository has been over
-          30s for more than 5% of requests during the last 5 minutes on cluster
-          {{ $labels.source_cluster }}
-        alert_routing_key: image-controller
-        runbook_url: https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/image-controller/latency_image_repository_provision.md
+    - name: latency-image-repository-provision
```
yftacherzog, line 9: Is the fact that the entire block changed here is expected?

test/promql/tests/data_plane/release_service_test.yaml
```diff
@@ -6,14 +6,22 @@ evaluation_interval: 1m
 tests:
   - interval: 1m
     input_series:
-      - series: 'release_processing_duration_seconds_bucket{le="3600", job="release", namespace="foo", source_cluster="cluster01"}'
```
yftacherzog, line 9: Is this change related to this PR?

yftacherzog (commented):
There are a few changes that looks like fixing of indentations etc. I would prefer this not to be done as part of such a big PR. There's enough potential for things to break as-is. Not sure we need to add additional changes along the way.

test/promql/tests/data_plane/release_service_test.yaml
Omeramsc: fixed

yftacherzog (approved):
lgtm

amisstea (approved):

PR #205, opened 2024-02-15
chore(RHTAPWATCH-741): Add recording rules for exporter

hmariset:
Add recording rules to translate the components existing metrics to expose availability.

PR #208, opened 2024-02-16
docs: Add recording rule documentation

hmariset:
Add recording rule documentation to the
README

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,30 @@ especially in the case in which the exporter is external to the code it's monito
 
 For more detailed documentation on [Availability exporters](https://gitlab.cee.redhat.com/konflux/docs/documentation/-/blob/main/o11y/monitoring/availability_exporters.md?ref_type=heads)
 
+## Recording Rules
+
+Recording rules allow us to precompute frequently needed or computationally expensive expressions
+and save their result as a new set of time series. Recording rules are the go-to approach for
+speeding up the performance of queries that take too long to return. When other teams want to go
+with their own metrics format for exporters they need to adapt to desired metric form by
+translating it using a recording rule. 
+
+These recording rules should be put in the [rhobs/recording folder](rhobs/recording/). 
+
+The standard format is single availability metric `konflux_up` with labels `service` and `check`.
+Each time series will have the service and check labels for the name of the originating service
+and availability check it performed, respectively. The metric konflux_up should return either 0
+or 1 based on the availability of the component/service. If the service is up then the
+metric should return 1 else 0.
+
+[Recording rule example](rhobs/recording/exporter_recording_rules.yml) provided here has
+below format
+  ```
+  grafana_ds_up(check=prometheus-appstudio-ds) -> konflux_up(service=grafana, check=prometheus-appstudio-ds)
+  ```
+
+For more detailed documentation on [recording rules](https://docs.google.com/document/d/1Y72T10JGuJaeyeNexmS_qTHfDB8uxxq0zERRRSOZegg/edit?usp=sharing)
+
 ## Support
 
 - Slack: [#forum-konflux-o11y](https://app.slack.com/client/E030G10V24F/C04FDFTF8EB)


hmariset:
EC checks are failing due to reference id needed instead of branch name for tekton task `run-kubelint`. This change is done in PR - https://github.com/redhat-appstudio/o11y/pull/205

README.md
```diff
@@ -144,6 +144,13 @@ especially in the case in which the exporter is external to the code it's monito
 
 For more detailed documentation on [Availability exporters](https://gitlab.cee.redhat.com/konflux/docs/documentation/-/blob/main/o11y/monitoring/availability_exporters.md?ref_type=heads)
 
+## Recording Rules
+
+Recording rules allow us to precompute frequently needed or computationally expensive expressions and save their result as a new set of time series. Recording rules are the go-to approach for speeding up the performance of queries that take too long to return. When other teams want to go with their own metrics format for exporters they need to adapt to desired metric form by translating it to the recording rules.
```
yftacherzog, line 149: please add linebreaks at 90-100 characters.

README.md
```diff
@@ -144,6 +144,13 @@ especially in the case in which the exporter is external to the code it's monito
 
 For more detailed documentation on [Availability exporters](https://gitlab.cee.redhat.com/konflux/docs/documentation/-/blob/main/o11y/monitoring/availability_exporters.md?ref_type=heads)
 
+## Recording Rules
+
+Recording rules allow us to precompute frequently needed or computationally expensive expressions and save their result as a new set of time series. Recording rules are the go-to approach for speeding up the performance of queries that take too long to return. When other teams want to go with their own metrics format for exporters they need to adapt to desired metric form by translating it to the recording rules. 
+These recording rules should be put in the [rhobs/recording folder](https://github.com/redhat-appstudio/o11y/tree/main/rhobs). 
+
+For more detailed documentation on [recording rules](https://docs.google.com/document/d/1Y72T10JGuJaeyeNexmS_qTHfDB8uxxq0zERRRSOZegg/edit?usp=sharing)
```
yftacherzog, line 152: I think we need to have here the requirements for the format that the resulting metric should have. i.e. metric name and the required labels.

It would make most sense to me to have those requirements defined in the same PR as the recording rule example, as then we can compare the requirement and the example and adjust them together.

hmariset: I added the requirement as comment in the recording rule now. Please check if it's sufficient or do I need to specify anywhere else.

README.md
hmariset: Updated

README.md
```diff
@@ -144,6 +144,18 @@ especially in the case in which the exporter is external to the code it's monito
 
 For more detailed documentation on [Availability exporters](https://gitlab.cee.redhat.com/konflux/docs/documentation/-/blob/main/o11y/monitoring/availability_exporters.md?ref_type=heads)
 
+## Recording Rules
+
+Recording rules allow us to precompute frequently needed or computationally expensive expressions
+and save their result as a new set of time series. Recording rules are the go-to approach for
+speeding up the performance of queries that take too long to return. When other teams want to go
+with their own metrics format for exporters they need to adapt to desired metric form by
+translating it to the recording rules.
```
amisstea, line 153:
```suggestion
translating it using a recording rule. 
```

README.md
yftacherzog: I think we need the requirements mentioned in some place other than the recording rule. I think this section of the readme file is appropriate.

README.md
```diff
@@ -144,6 +144,23 @@ especially in the case in which the exporter is external to the code it's monito
...
+and save their result as a new set of time series. Recording rules are the go-to approach for
+speeding up the performance of queries that take too long to return. When other teams want to go
+with their own metrics format for exporters they need to adapt to desired metric form by
+translating it using a recording rule. 
+
+These recording rules should be put in the [rhobs/recording folder](https://github.com/redhat-appstudio/o11y/tree/main/rhobs). 
+
+Below is the metric we need from recording rules
+ ```
+ grafana_ds_up(check=prometheus-appstudio-ds) -> konflux_up(service=grafana, check=prometheus-appstudio-ds)
```
yftacherzog, line 159: We cannot anticipate the format of the original metric, so we need to provide clear requirements for the availability metric that we want to have eventually. I.e.

- Which labels the metric should have?
- What should those labels represent?
- What value of the metrics means up and what means down?
- We should require the metric to provide value both when it's up and when it's down (we know of metrics that only report when up - they are hard to work with).

We can mention a supporting example, but the requirements should be clear regardless.

hmariset: Updated

README.md
```diff
@@ -144,6 +144,30 @@ especially in the case in which the exporter is external to the code it's monito
...
+translating it using a recording rule. 
+
+These recording rules should be put in the [rhobs/recording folder](https://github.com/redhat-appstudio/o11y/tree/main/rhobs). 
+
+The standard format is single availability metric konflux_up with labels service and check.
+Each time series will have the service and check labels for the name of the originating service
+and availability check it performed, respectively. The metric konflux_up will return either 0
+or 1 based on the availability of the component/service. For example exporter which we are
+checking it returns 1 if prometheus-appstudio-ds datasource exists in our grafana service else
+it returns 0.
```
amisstea, line 162:
```suggestion
or 1 based on the availability of the component/service. In the case of the example exporter,
it returns 1 if prometheus-appstudio-ds datasource exists in our grafana service else it returns 0.
```

README.md
```diff
@@ -144,6 +144,30 @@ especially in the case in which the exporter is external to the code it's monito
...
+
+Recording rules allow us to precompute frequently needed or computationally expensive expressions
+and save their result as a new set of time series. Recording rules are the go-to approach for
+speeding up the performance of queries that take too long to return. When other teams want to go
+with their own metrics format for exporters they need to adapt to desired metric form by
+translating it using a recording rule. 
+
+These recording rules should be put in the [rhobs/recording folder](https://github.com/redhat-appstudio/o11y/tree/main/rhobs). 
+
+The standard format is single availability metric konflux_up with labels service and check.
```
amisstea, line 157:
```suggestion
The standard format is single availability metric `konflux_up` with labels `service` and `check`.
```

README.md
hmariset: Updated

README.md
hmariset: Updated

hmariset:
@amisstea Please review again

amisstea (approved):

README.md
```diff
@@ -144,6 +144,29 @@ especially in the case in which the exporter is external to the code it's monito
...
+with their own metrics format for exporters they need to adapt to desired metric form by
+translating it using a recording rule. 
+
+These recording rules should be put in the [rhobs/recording folder](https://github.com/redhat-appstudio/o11y/tree/main/rhobs). 
+
+The standard format is single availability metric `konflux_up` with labels `service` and `check`.
+Each time series will have the service and check labels for the name of the originating service
+and availability check it performed, respectively. The metric konflux_up will return either 0
+or 1 based on the availability of the component/service. In the case of the example exporter,
+it returns 1 if prometheus-appstudio-ds datasource exists in our grafana service else it returns 0.
```
yftacherzog, line 161: It's a bit ambiguous here. We should state it as a requirement that 1 should mean available and 0 should mean unavailable.

hmariset: @yftacherzog updated it, Please check.

README.md
```diff
@@ -146,6 +146,29 @@ especially in the case in which the exporter is external to the code it's monito
...
 
+## Recording Rules
+
+Recording rules allow us to precompute frequently needed or computationally expensive expressions
+and save their result as a new set of time series. Recording rules are the go-to approach for
+speeding up the performance of queries that take too long to return. When other teams want to go
+with their own metrics format for exporters they need to adapt to desired metric form by
+translating it using a recording rule. 
+
+These recording rules should be put in the [rhobs/recording folder](https://github.com/redhat-appstudio/o11y/tree/main/rhobs).
```
yftacherzog, line 157: The link is not pointing to the location mentioned in the link's text.
Also, you can use relative path for links inside the same repo.
See this example: https://stackoverflow.com/questions/7653483/github-relative-link-in-markdown-file

README.md
```diff
@@ -146,6 +146,29 @@ especially in the case in which the exporter is external to the code it's monito
...
+
+These recording rules should be put in the [rhobs/recording folder](https://github.com/redhat-appstudio/o11y/tree/main/rhobs). 
+
+The standard format is single availability metric `konflux_up` with labels `service` and `check`.
+Each time series will have the service and check labels for the name of the originating service
+and availability check it performed, respectively. The metric konflux_up should return either 0
+or 1 based on the availability of the component/service. If the service is up then the metric
+should return 1 else 0.
+
+Recording rule example provided here has below format
```
yftacherzog, line 165: We should include the link to the reference rule.

README.md
hmariset: Thanks @yftacherzog! Updated the relative link here

README.md
hmariset: Added the link for reference rule.

yftacherzog (approved):
lgtm

PR #209, opened 2024-02-18
fix: use sha instead of branch name for clone-repo step

yftacherzog:
Using branch name causes EC to fail. Using commit sha instead


avi-biton (approved):
/lgtm

PR #225, opened 2024-03-04
chore: Update cluster role name

hmariset:
Updated cluster role name since it's conflicting
with other ArgoCD components resource name.


amisstea (commented):
This seems ok but can you clarify which other component is deploying the `kube-rbac-proxy` ClusterRole to the `appstudio-grafana-datasource-exporter` namespace?

hmariset:
> This seems ok but can you clarify which other component is deploying the `kube-rbac-proxy` ClusterRole to the `appstudio-grafana-datasource-exporter` namespace?

It's not deploying to the same namespace but we are getting warning that `kube-rbac-proxy` is used between `monitoring-workload-prometheus-in-cluster-local` and `monitoring-workload-grafana-in-cluster-local`

amisstea (approved):

PR #228, opened 2024-03-05
fix(RHTAPSRE-437): Removed Child Folder from dashboards folder

manish-jangra:
This change removes the child folder and keeps the ArgoCD Dashboard in the same parent folder `/dashboards`
Signed-off-by: Manish Kumar


avi-biton (approved):
LGTM

PR #238, opened 2024-03-19
fix(RHTAPSRE-485): Removing namespace regex

manish-jangra:
removing namespace regex

Signed-off-by: Manish Kumar


avi-biton (approved):
lgtm

PR #242, opened 2024-03-24
chore(RHTAPWATCH-794): update alert_routing_key in argocd alerts

avi-biton:
Update `alert_routing_key` to be assigned with the application name instead of namespace in ArgoCD alerts


yftacherzog (approved):
lgtm

PR #245, opened 2024-03-28
feat(STONE-855): making alerts less sensitive

jsztuka:
Increasing alert **for** value to make trigering of alerts less sensitive.

### DIFF
--- a/rhobs/alerting/data_plane/prometheus.latency_pac_provision_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.latency_pac_provision_alerts.yaml
@@ -19,7 +19,7 @@ spec:
           ) / ignoring(le)
           increase(redhat_appstudio_buildservice_PaC_configuration_time_bucket{le="+Inf"}[5m])
         ) > 0.05
-      for: 1m
+      for: 5m
       labels:
         severity: critical
         slo: "true"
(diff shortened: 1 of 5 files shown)


dirgim:
I think these changes will also require updates to unit tests at https://github.com/redhat-appstudio/o11y/tree/main/test/promql/tests/data_plane

rhobs/alerting/data_plane/prometheus.latency_pac_provision_alerts.yaml
```diff
@@ -19,7 +19,7 @@ spec:
           ) / ignoring(le)
           increase(redhat_appstudio_buildservice_PaC_configuration_time_bucket{le="+Inf"}[5m])
         ) > 0.05
-      for: 1m
+      for: 10m
```
dirgim, line 22: I don't think this is an alert owned by the integration service, if we want to update it, we should probably notify the build service team about it.

jsztuka:
@mmorhun 
https://github.com/redhat-appstudio/o11y/blob/main/rhobs/alerting/data_plane/prometheus.latency_pac_provision_alerts.yaml
Could be running into same troubles as other two mentioned alerts, do you want me to update it also?

dirgim (approved):
lgtm, but will defer to the o11y team

amisstea (approved):
LGTM

mmorhun:
@jsztuka we haven't experienced problems so far with PaC metrics, but I think we can increase the value, because 1 minute looks too sensetive

yftacherzog (approved):

PR #248, opened 2024-04-09
chore(RHTAPWATCH-890): Add Alering Rule for konflux_up metric

hmariset:
Added alerting rule and related promql unit tests
for konflux_up metric.

Jira-Url:https://issues.redhat.com/browse/RHTAPWATCH-890

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,11 @@ The different alerting rules in this repository are:
 
 * [**Alert Rule QuotaExceeded**](https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/main/o11y/alert-rule-QuotaExceeded.md)
 
+### Availability Metric Alerts
+
+These Alert rules are defined to monitor and alert if the `konflux_up` metric is missing
+for all expected permutations of the `service` and `check` labels across different environments.
+
 ### SLO Alerts
 
 SLO (Service Level Objective) alert rules are rules defined to monitor and alert 
--- a/rhobs/alerting/data_plane/prometheus.availability_metric_grafana_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.availability_metric_grafana_alerts.yaml
@@ -1,0 +1,22 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: rhtap-konflux-up-grafana-alerting
+  labels:
+    tenant: rhtap
+spec:
+  groups:
+  - name: konflux_up_grafana_alerts
+    interval: 1m
+    rules:
+    - alert: KonfluxUpGrafanaAlert
+      expr: absent(konflux_up{service="grafana", check="prometheus-appstudio-ds"}) == 1
+      for: 5m
+      labels:
+        severity: warning
+      annotations:
+        summary: Availability metric 'konflux_up' is missing.
+        description: >-
+          Availability metric 'konflux_up' is missing for check {{ $labels.check }}
+          on service {{ $labels.service }} for more than 5 minutes.
+        alert_routing_key: 'appstudio-grafana'
--- a/rhobs/alerting/data_plane/prometheus.availability_metric_has_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.availability_metric_has_alerts.yaml
@@ -1,0 +1,22 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: rhtap-konflux-up-has-alerting
+  labels:
+    tenant: rhtap
+spec:
+  groups:
+  - name: konflux_up_has_alerts
+    interval: 1m
+    rules:
+    - alert: KonfluxUpHasAlert
+      expr: absent(konflux_up{service="has", check="github"}) == 1
+      for: 5m
+      labels:
+        severity: warning
+      annotations:
+        summary: Availability metric 'konflux_up' is missing.
+        description: >-
+          Availability metric 'konflux_up' is missing for check {{ $labels.check }}
+          on service {{ $labels.service }} for more than 5 minutes.
+        alert_routing_key: 'appstudio-grafana'
--- a/rhobs/alerting/data_plane/prometheus.availability_metric_remotesecret_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.availability_metric_remotesecret_alerts.yaml
@@ -1,0 +1,22 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: rhtap-konflux-up-remote-secret-alerting
+  labels:
+    tenant: rhtap
+spec:
+  groups:
+  - name: konflux_up_remote_secret_alerts
+    interval: 1m
+    rules:
+    - alert: KonfluxUpRemoteSecretAlert
+      expr: absent(konflux_up{service="remote-secret-controller-manager-metrics-service", check="secretstorage"}) == 1
+      for: 5m
+      labels:
+        severity: warning
+      annotations:
+        summary: Availability metric 'konflux_up' is missing.
+        description: >-
+          Availability metric 'konflux_up' is missing for check {{ $labels.check }}
+          on service {{ $labels.service }} for more than 5 minutes.
+        alert_routing_key: 'appstudio-grafana'
--- a/test/promql/tests/data_plane/availability_metric_grafana_test.yaml
+++ b/test/promql/tests/data_plane/availability_metric_grafana_test.yaml
@@ -1,0 +1,36 @@
+evaluation_interval: 1m
+
+rule_files:
+  - prometheus.availability_metric_grafana_alerts.yaml
+
+tests:
+  - interval: 1m
+    input_series:
+      - series: 'konflux_up{service="grafana", check="prometheus-appstudio-ds"}'
+        values: '1x15'
+
+    alert_rule_test:
+      - eval_time: 15m
+        alertname: KonfluxUpGrafanaAlert
+
+  - interval: 1m
+    input_series:
+      - series: 'konflux_up{service="has", check="github"}'
+        values: '_x15'
+
+    alert_rule_test:
+
+      - eval_time: 15m
+        alertname: KonfluxUpGrafanaAlert
+        exp_alerts:
+          - exp_labels:
+              severity: warning
+              check: prometheus-appstudio-ds
+              service: grafana
+            exp_annotations:
+              summary: >-
+                Availability metric 'konflux_up' is missing.
+              description: >-
+                Availability metric 'konflux_up' is missing for check prometheus-appstudio-ds
+                on service grafana for more than 5 minutes.
+              alert_routing_key: 'appstudio-grafana'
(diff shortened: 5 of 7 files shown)


rhobs/alerting/data_plane/prometheus.exporter_alerts.yaml
```diff
@@ -0,0 +1,49 @@
...
+  name: konflux-exporter-alerting
+  labels:
+    tenant: rhtap
+spec:
+  groups:
+  - name: exporter_alerts
+    interval: 5m
+    rules:
+    - alert: KonfluxUpAlert
+      expr: absent_over_time(konflux_up{namespace="appstudio-grafana-datasource-exporter"}[5m]) == 1
```
yftacherzog, line 13: I think we should also add the `check` label to the expression here, so if we add another konflux_up metric on that namespace in the future then we won't get false negatives.

yftacherzog: Can we use `absent` instead of `absent_over_time`?
Using queries that look at a range vector (e.g. `[5m]`) makes it less trivial to understand when they are supposed to alert, comparing to metrics looking at instant vector (no `[]`).

rhobs/alerting/data_plane/prometheus.exporter_alerts.yaml
```diff
@@ -0,0 +1,49 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: konflux-exporter-alerting
+  labels:
+    tenant: rhtap
+spec:
+  groups:
+  - name: exporter_alerts
+    interval: 5m
```
yftacherzog, line 10: I don't think we should use non-default interval here. It will needlessly make it more complex.

rhobs/alerting/data_plane/prometheus.exporter_alerts.yaml
```diff
@@ -0,0 +1,49 @@
...
+      expr: absent_over_time(konflux_up{namespace="appstudio-grafana-datasource-exporter"}[5m]) == 1
+      for: 5m
+      labels:
+        severity: warning
+      annotations:
+        summary: Exporter metric 'konflux_up' is missing.
+        description: >-
+          The Prometheus exporter metric 'konflux_up' for namespace
+          "appstudio-grafana-datasource-exporter" on cluster
+          {{ $labels.source_cluster }} is missing for more than 30 minutes.
```
yftacherzog, line 22: It was 5 minutes window for 5 minutes, I think it should not mean 30 minutes. Using `absent` should make it simpler to understand the time window.

rhobs/alerting/data_plane/prometheus.exporter_alerts.yaml
yftacherzog, line 1: The exporter is what we're using to generate the metric that `konflux_up` is eventually generated from. Others may generate `konflux_up` from something else, so those alerts are more about the presence of the availability metric than the exporter.

Maybe we can call it `prometheus.availability_metric_alerts.yaml` instead.

rhobs/alerting/data_plane/prometheus.exporter_alerts.yaml
```diff
@@ -0,0 +1,49 @@
...
+        alert_routing_key: 'appstudio-grafana-datasource-exporter'
```
yftacherzog, line 23: @avi-biton would that mean creating a new entry for routing this new key?
Maybe we can use our existing key for routing for now.

rhobs/alerting/data_plane/prometheus.exporter_alerts.yaml
```diff
@@ -0,0 +1,49 @@
...
+  - name: exporter_alerts
+    interval: 5m
+    rules:
+    - alert: KonfluxUpAlert
+      expr: absent_over_time(konflux_up{namespace="appstudio-grafana-datasource-exporter"}[5m]) == 1
+      for: 5m
+      labels:
+        severity: warning
+      annotations:
+        summary: Exporter metric 'konflux_up' is missing.
```
yftacherzog, line 18:
```suggestion
        summary: Availability metric 'konflux_up' is missing.
```

rhobs/alerting/data_plane/prometheus.exporter_alerts.yaml
```diff
@@ -0,0 +1,49 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: konflux-exporter-alerting
```
yftacherzog, line 4:
```suggestion
  name: konflux-up-alerting
```

rhobs/alerting/data_plane/prometheus.exporter_alerts.yaml
```diff
@@ -0,0 +1,49 @@
...
+  labels:
+    tenant: rhtap
+spec:
+  groups:
+  - name: exporter_alerts
```
yftacherzog, line 9:
```suggestion
  - name: konflux_up_alerts
```

test/promql/tests/data_plane/exporter_test.yaml
```diff
@@ -0,0 +1,64 @@
...
+rule_files:
+  - prometheus.exporter_alerts.yaml
+
+tests:
+  - interval: 1m
+    input_series:
+      - series: 'konflux_up{namespace="application-service", source_cluster="cluster01"}'
+        values: '1x15'
+      - series: 'metric_up{namespace="some-service", source_cluster="cluster01"}'
+        values: '0x15'
```
yftacherzog, line 12: What's the reason for adding this time series?

test/promql/tests/data_plane/exporter_test.yaml
```diff
@@ -0,0 +1,64 @@
...
+          - exp_labels:
+              severity: warning
+              namespace: appstudio-grafana-datasource-exporter
+            exp_annotations:
+              summary: >-
+                Exporter metric 'konflux_up' is missing.
+              description: >-
+                The Prometheus exporter metric 'konflux_up' for namespace
+                "appstudio-grafana-datasource-exporter" on cluster  
+                is missing for more than 30 minutes.
```
yftacherzog, line 27: The cluster label is mentioned on the rule, but you don't get it printed here as there's no time series to take labels from.
This actually raises an issue here. Those alert rules are deployed in RHOBS, where we have data coming from all clusters. It means that this alert will only fire if the metric is missing on all clusters, which is not what we want (we want it to alert as soon as it's missing from any cluster).

Try thinking if there's something we can do about it (I don't have a solution at the moment).

yftacherzog (changes requested):
As it is at the moment, the alert will fire only when the metric is missing from all clusters. We need to find a way to alert when it's missing from any cluster.

test/promql/tests/data_plane/exporter_test.yaml
yftacherzog: (Thinking out loud)
Suppose that we somehow configure a recording rule on each cluster with the absence metric computed for the any combinations of labels for the konflux_up metric there (e.g. `absence_konflux_up`), then we should get a reading for that recording rule per-check&service, per-cluster once `konflux_up` is missing for any such combination on any cluster.

In that case, we can have a single alerting rule in RHOBS that will alert as soon as any time series in `absence_konflux_up` is giving a reading of `1`.

I think it should work in theory, but we don't have a mechanism in place for defining recording rules on clusters themselves.

rhobs/alerting/data_plane/prometheus.exporter_alerts.yaml
amisstea: Looking at the alertmanager [config](https://gitlab.cee.redhat.com/service/app-interface/-/blob/d35086d821680821d3f6b9827e41b45a6bd16da8/resources/observability/alertmanager/alertmanager-instance.secret.yaml#L3555), there's no condition with this routing key but there is one for `appstudio-grafana`. I think we can use that instead.

rhobs/alerting/data_plane/prometheus.availability_metric_alerts.yaml
```diff
@@ -0,0 +1,45 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: konflux-up-alerting
```
amisstea, line 4:
```suggestion
  name: rhtap-konflux-up-alerting
```
or
```suggestion
  name: rhtap-up-alerting
```
Our alerting rules still need to be prefixed with `rhtap-` to pass the CI pipeline. The first option seems slightly less terrible to me.

amisstea (changes requested):

test/promql/tests/data_plane/availability_metric_test.yaml
yftacherzog, line 1: I think we should have 2 test cases per each rule. One in which the time series is there and another in which it's missing (not-alerting/alerting). Currently, they are all copy-paste of each other, but suppose you change `KonfluxUpGrafanaAlert` tomorrow, there's no test case covering its regressions.

hmariset:
@yftacherzog If interval is not specified then the `check-and-test-prometheus-rules` is failing.

test/promql/tests/data_plane/availability_metric_test.yaml
hmariset: Added additional test cases.

README.md
```diff
@@ -31,6 +31,10 @@ The different alerting rules in this repository are:
 
 * [**Alert Rule QuotaExceeded**](https://gitlab.cee.redhat.com/rhtap/docs/sop/-/blob/main/o11y/alert-rule-QuotaExceeded.md)
 
+### Availability Metric Alerts
+
+These Alert rules are defined to monitor and alert if the konflux_up metric is available for Observability, SPI and HAS teams for different environments.
```
amisstea, line 30:
```suggestion
These Alert rules are defined to monitor and alert if the `konflux_up` metric is missing for all expected permutations of the `service` and `check` labels across different environments.
```

amisstea (approved):
The alerting rules LGTM. Just one suggestion about the doc update.

yftacherzog:
> @yftacherzog If interval is not specified then the `check-and-test-prometheus-rules` is failing.

So let's make it a 1m as the rest of the rules we have.

hmariset:
> > @yftacherzog If interval is not specified then the `check-and-test-prometheus-rules` is failing.
> 
> So let's make it a 1m as the rest of the rules we have.

@yftacherzog, Already updated to 1m

README.md
```diff
@@ -25,6 +25,10 @@ The different alerting rules in this repository are:
 
 * [**Alert Rule QuotaExceeded**](https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/main/o11y/alert-rule-QuotaExceeded.md)
 
+### Availability Metric Alerts
+
+These Alert rules are defined to monitor and alert if the `konflux_up` metric is missing for all expected permutations of the `service` and `check` labels across different environments.
```
yftacherzog, line 30: break this line at 80-90 chars

rhobs/alerting/data_plane/prometheus.availability_metric_alerts.yaml
yftacherzog, line 1: If I understand this correctly, once you have all the alerts defined on the same file, and given the nature of the absent function (it alerts for anything that is not there), you practically need to test them all together. This means that if tomorrow we add another alert rule to this file, we'd have to change all the test cases we have in place.

If this is indeed the case, then I think we should have each alert rule defined and tested in files of its own.

hmariset:
@amisstea @yftacherzog Placed the rules in different alert files. Please check again.

yftacherzog (approved):
lgtm

PR #267, opened 2024-05-16
fix(KFLUXBUGS-1248): adjust panel threshold to match query

gabemontero:
since we went from <query> to "1 - <query>" we had to flip the threshold base and colors so that it is green if 95 or above and red if below 95

rh-pre-commit.version: 2.2.0


gabemontero:
@amisstea PTAL

amisstea (approved):

PR #273, opened 2024-06-04
fix(RHTAPWATCHO-1052): update glibc because of CVE vulnerabilities

Omeramsc:
There seem to be an issue with 3 high level CVE vulnerabilities, related to glibc.

This update upgrade glibc manually as the parent image does not contain the fix yet.


yftacherzog (approved):
lgtm. I prefer this approach comparing to the one with dnf upgrade. Hopefully no automation will try to downgrade it later.

PR #282, opened 2024-06-21
chore(build team SLOs) Correct Build Team SLO alerts

mmorhun:
This PR changes the following SLO alerts:
- Build Service simple build: since simple builds are not often used, even one failure would trigger alert. Increasing thresholds to be less sensitive.
- Image Controller: sometimes Quay stability is not ideal and we shouldn't alert at the first failure. Increasing threshold to 20% since the operation is not often performed (the same for the time period).


avi-biton:
LGTM

mmorhun:
I don't have permissions to merge, could someone click the button?

PR #295, opened 2024-07-12
feat(RELEASE-996): availability metrics for release-service

theflockers:
This commit adds availability record rules and tests for the release-service availability metrics.

### DIFF
--- a/rhobs/recording/release_service_availability_recording_rules.yaml
+++ b/rhobs/recording/release_service_availability_recording_rules.yaml
@@ -1,0 +1,20 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: rhtap-release-service-availability
+  labels:
+    tenant: rhtap
+spec:
+
+  groups:
+    - name: release_service_general_availability
+      interval: 1m
+      rules:
+        - record: konflux_up
+          expr: |
+            label_replace(
+              avg(release_service_check_gauge) without(Name) OR
+              label_replace(
+                floor(sum(release_service_check_gauge) without(check) / count(release_service_check_gauge) without(check)),
+              "check", "release", "", ""),
+            "Name", "release_service_check_gauge", "", "")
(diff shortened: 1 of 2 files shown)


amisstea:
@theflockers the EC failures should be resolved by rebasing your branch against `main`.

rhobs/recording/release_service_availability_recording_rules.yaml
```diff
@@ -0,0 +1,20 @@
...
+spec:
+
+  groups:
+    - name: release_service_general_availability
+      interval: 1m
+      rules:
+        - record: konflux_up
+          expr: |
+            label_replace(
+              avg(release_service_check_gauge) without(Name) OR
```
yftacherzog, line 16: As we mention the label `Name` here, I think we should also have it inside the input series to verify things are aggregated as expected.

theflockers: True. As I am artificially adding the "release"  one, it duplicates the Name - hence I had to remove it, before re-adding it. But I forgot to update with the last operation. Fixed!

theflockers:
The PR got weird after the rebase. I will close and recreate.

theflockers:
> The PR got weird after the rebase. I will close and recreate.

could fix it!!

yftacherzog:
> > The PR got weird after the rebase. I will close and recreate.
> 
> Less bad. I could fix it!!

I can't see the change you mentioned though :)

theflockers:
> > > The PR got weird after the rebase. I will close and recreate.
> > 
> > 
> > Less bad. I could fix it!!
> 
> I can't see the change you mentioned though :)

Yeah. I had to force push it: https://github.com/redhat-appstudio/o11y/blob/9ee0c5419b8bd63ae54333fb6134bd5dca7558aa/rhobs/recording/release_service_availability_recording_rules.yaml#L20

Anyhow, something else is breaking now. Checking it.

rhobs/recording/release_service_availability_recording_rules.yaml
yftacherzog: The test now shows that it's being added. I think we should also have it in some of the inputs to test how we handle also cases in which it's present in the first place in `release_service_check_gauge`.

yftacherzog (approved):
lgtm

theflockers:
@yftacherzog please merge this one whenever you can. Thanks!

PR #297, opened 2024-07-15
WIP: SRVKP-4523: add alert for tekton-pipeline-controller crashlooping

gabemontero:
I don't have the tests and panels yet (panels will look close to what I have for the pipeline service dashboard at https://github.com/openshift-pipelines/pipeline-service/pull/1020/files#diff-13716f85d071b7d6b050dd4f5d738d85f4627f17f20ee8b68fd8f39084c72ce9R229-R292 ) but I wanted to get the alert query up for some early feedback in https://issues.redhat.com/browse/KONFLUX-3579 

rh-pre-commit.version: 2.3.0

### DIFF
--- a/rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
@@ -47,3 +47,18 @@ spec:
             alert_team_handle: <!subteam^S03GF42RBE2>
             team: pipelines
             runbook_url: TBD
+        - alert: CorePipelineControllerCrashloop
+          expr: |
+            sum(kube_pod_container_status_restarts_total{namespace="openshift-pipelines", pod=~"tekton-pipelines-controller-.*"}) - sum(kube_pod_container_status_restarts_total{namespace="openshift-pipelines", pod=~"tekton-pipelines-controller-.*"} offset 2m) > 0
+          for: 5m
+          labels:
+            severity: critical
+            slo: "true"
+          annotations:
+            summary: >-
+              Tekton controller is crashlooping
+            description: >-
+              Tekton controller on cluster {{ $labels.source_cluster }} has restarted {{ $value }} the past 5 minutes.
+            alert_team_handle: <!subteam^S03GF42RBE2>
+            team: pipelines
+            runbook_url: TBD


rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
```diff
@@ -47,3 +47,18 @@ spec:
             alert_team_handle: <!subteam^S03GF42RBE2>
             team: pipelines
             runbook_url: TBD
+        - alert: CorePipelineControllerCrashloop
+          expr: |
+            sum(kube_pod_container_status_restarts_total{namespace="openshift-pipelines", pod=~"tekton-pipelines-controller-.*"}) - sum(kube_pod_container_status_restarts_total{namespace="openshift-pipelines", pod=~"tekton-pipelines-controller-.*"} offset 2m) > 0
```
zanssa, line 52: Could I ask why we are calculating the difference between the current total of restart and the total of restart over 2 mins?  My experience is humble so I am wondering about the expected behavior when let's say the current total of restart is 0 but the total of restart over the past 2 mins is 4. The result is not greater than 0 but it's negative. So, there will be no alert. correct or would a negative result cause an issue in showing data? 


Also, what do you think about using rate function: https://prometheus.io/docs/prometheus/latest/querying/functions/#rate ? If we are interested in monitoring the total number of restart ( increase over time / decrease ) wouldn't this do the job and simplify the query? 

for example, `rate(kube_pod_container_status_restarts_total{namespace="openshift-pipelines", pod=~"tekton-pipelines-controller-.*"}[5m]) >0`


If a restart happens because let's say we are doing an upgrade for the operator, the alert shouldn't be triggered. It should be triggered when the pos restarts several times over X period.

rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
```diff
@@ -47,3 +47,18 @@ spec:
...
+          for: 5m
+          labels:
+            severity: critical
+            slo: "true"
+          annotations:
+            summary: >-
+              Tekton controller is crashlooping
```
zanssa, line 59: The metric is `kube_pod_container_status_restarts_total` and the restart could happen for another reason than crashlooping. For example, the restart could be due to mem issue, OOMKilled. So, we might need to reword this. thoughts?

gabemontero: you are correct here @zanssa I will reword

rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
gabemontero: you are correct here as well I believe based on your comment @zanssa .  Aside from `rate(...)` I know remember @yftacherzog also coaching me on using `increase(...)` to further deal with counter resets from the monitoring system. 

I will make that update as well.

gabemontero: check that after some tries in the metric console and some internet searching i.e. https://stackoverflow.com/questions/54494394/do-i-understand-prometheuss-rate-vs-increase-functions-correctly I think I need to use `increase(...)` as a slightly augmented alternative to `rate(...)`

PR #298, opened 2024-07-16
fix: reduce sensitivity of the PodsNotReady alerting rule

amisstea:
Many of these alerts are auto resolving shortly after the 15m window. Increasing this to 30m in an effort to reduce alert fatigue.


amisstea:
A fix for the EC failures is available in #299.

yftacherzog (approved):
lgtm

PR #299, opened 2024-07-16
fix: use OCI trusted artifacts in CI

amisstea:
There are a number of untrusted tasks in the CI pipeline that trigger failures in the EC integration pipeline. The EnterpriseContractPolicy requires all tasks to be trusted unless operating with OCI trusted artifacts. When using trusted artifacts, EC should not raise violations for tasks which are not involved in creating an artifact.


yftacherzog (approved):
lgtm. Thank you for looking into that!

PR #317, opened 2024-07-25
chore(SRVKP-4532): expand tekton restart alert

gabemontero:

expanding the restart alert to handle all `tekton-` pods in `openshift-namespace` given the recent trouble with the remote-resolver deployment 

we'll handle PAC and Results (especially given the memory leak) separately

@amisstea PTAL

@gbenhaim @zanssa @enarha  @divyansh42 @savitaashture FYI

rh-pre-commit.version: 2.3.0

### DIFF
--- a/dashboards/grafana-dashboard-rhtap-slos.configmap.yaml
+++ b/dashboards/grafana-dashboard-rhtap-slos.configmap.yaml
@@ -1261,7 +1261,7 @@ data:
             "type": "prometheus",
             "uid": "${datasource}"
           },
-          "description": "The number of times the pipelines controller has restarted",
+          "description": "The number of times any of the tekton controllers have restarted",
           "fieldConfig": {
             "defaults": {
               "color": {
@@ -1328,7 +1328,7 @@ data:
                 "uid": "${datasource}"
               },
               "editorMode": "code",
-              "expr": "sum(increase(kube_pod_container_status_restarts_total{namespace=\"openshift-pipelines\", pod=~\"tekton-pipelines-controller-.*\"}[$__range]))",
+              "expr": "sum(increase(kube_pod_container_status_restarts_total{namespace=\"openshift-pipelines\", pod=~\"tekton-.*\"}[$__range]))",
               "instant": false,
               "legendFormat": "__auto",
               "range": true,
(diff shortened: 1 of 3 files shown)


amisstea (approved):

gabemontero:
confirmed this is up on staging

even caught the recent rash of restarts when you did a 28 day query

narrow to the last day and back to 0

@zanssa FYI

zanssa:
> confirmed this is up on staging
> 
> even caught the recent rash of restarts when you did a 28 day query
> 
> narrow to the last day and back to 0
> 
> @zanssa FYI

Hmm. Thanks for keeping me in the loop, Gabe. I looked at the panel in grafana instance (I used the same link you mentioned in the other PR) and if I specified the time range to be only June 27 (the day we had the incident) or even if I set the timeframe between June 26 to 28,  I wouldn't see data in the panel that indicates there was a restart. What I see is N/A. Am I missing something?

gabemontero:
> > confirmed this is up on staging
> > even caught the recent rash of restarts when you did a 28 day query
> > narrow to the last day and back to 0
> > @zanssa FYI
> 
> Hmm. Thanks for keeping me in the loop, Gabe. I looked at the panel in grafana instance (I used the same link you mentioned in the other PR) and if I specified the time range to be only June 27 (the day we had the incident) or even if I set the timeframe between June 26 to 28, I wouldn't see data in the panel that indicates there was a restart. What I see is N/A. Am I missing something?

so @zanssa 

The last 24 hour query https://grafana.stage.devshift.net/d/rhtap-slos/rhtap-slos?orgId=1&refresh=1h&from=now-24h&to=now produces 

restarts-24-hours

with no restarts

Last 30 days as of today July 31  i.e. https://grafana.stage.devshift.net/d/rhtap-slos/rhtap-slos?orgId=1&refresh=1h&from=now-30d&to=now still produces the high number of restarts that occured a week or so ago

last-30-day-restarts

As to why you can't see stuff from June, that is because the metric for restarts had not been added to the approved list os what could be queried.  I did that only 2 weeks ago, as part of my efforts here.  The change was https://github.com/redhat-appstudio/infra-deployments/pull/4105/files#diff-0a3efc78b1b7e372f5fc3f14b25f37861c74a0a9dae1522635051d00caeef8d7R105

PR #335, opened 2024-08-14
fix(RELEASE-1031): remove release_processing_duration SLO

theflockers:
This commit removes the release_processing_duration SLO from alerting and dashboard as for the current moment the metrics available in this sensor can not be trusted as releases can have different processing time for different users.


johnbieren (approved):

yftacherzog (approved):
lgtm

PR #360, opened 2024-08-29
fix: post merge pipeline svc panel tweaks

gabemontero:

Per discussions with @amisstea @divyansh42 here is the panel update which moves the slo template back to the bottom, and fixes some of the pipeline service panel groupings after all the (relatively) concurrently panel merges between my PRs and @divyansh42 's

Also
- there was a regexp tweak to the restart query that got list with @amisstea 's rebase of @divyansh42 's PR ... put that back in
- some of the results go to read threshold values were more aggressive than what we have in the alerts themselves
- also forgot that the default range is 28 days for the panels, where the new alerts or more in the 75 minute window; as a result, history from using `sum(increase(...)` that we were guided to doing for the alerts actually paint a more negative picture that what current reality when viewed over 28 days for some of the queries.... so I switched the display to `sum(rate(...))` and the values there were more in line with what I see when I do a `sum(increase(...))` ... over an hour duration, so I pivoted to that and adjusted some of the verbiage
- lastly, per the Aug 21 arch call, following the pattern other teams talked about as part of leaving the panels for alerts we removed because they required infra-deployment repo changes, I retitled those panels so it was clear they were not metrics we alert on .... if need be down the line we can remove them if there are objections
  
lastly the screen shots (needed 2 even with my 4K monitor we have so many now)

screenshot-1

screenshot-2

rh-pre-commit.version: 2.3.1

### DIFF
--- a/dashboards/grafana-dashboard-rhtap-slos.configmap.yaml
+++ b/dashboards/grafana-dashboard-rhtap-slos.configmap.yaml
@@ -662,12 +662,1532 @@ data:
                 "type": "gauge"
             },
             {
-                "collapsed": true,
+                "collapsed": false,
                 "gridPos": {
                     "h": 1,
                     "w": 24,
                     "x": 0,
                     "y": 40
+                },
+                "id": 64,
+                "panels": [],
+                "title": "Pipeline SLOs",
+                "type": "row"
+            },
+            {
+                "datasource": {
+                    "type": "prometheus",
+                    "uid": "${datasource}"
+                },
+                "description": "Factoring out the time needed to receive PipelineRun creation events from the overall PipelineRun execution time is at or above the defined percentile",
+                "gridPos": {
+                    "h": 7,
+                    "w": 10,
+                    "x": 0,
+                    "y": 41
+                },
+                "id": 65,
+                "options": {
+                    "code": {
+                        "language": "plaintext",
+                        "showLineNumbers": false,
+                        "showMiniMap": false
+                    },
+                    "content": "<h1><center>95% of pipeline run duration excludes scheduling overhead</center></h1>",
+                    "mode": "html"
+                },
+                "pluginVersion": "10.4.1",
+                "targets": [
+                    {
+                        "datasource": {
+                            "type": "prometheus",
+                            "uid": "${datasource}"
+                        },
+                        "refId": "A"
+                    }
+                ],
+                "title": "Pipeline Scheduling Overhead (alert removed, panel removal occurring soon)",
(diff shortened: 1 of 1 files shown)


dashboards/grafana-dashboard-rhtap-slos.configmap.yaml
```diff
@@ -905,7 +706,7 @@ data:
                         "refId": "A"
                     }
                 ],
-                "title": "Pipeline Scheduling SLO",
+                "title": "Pipeline Scheduling Overhead (not an SLO or alertable metric)",
```
amisstea, line 709: Maybe we should start a new pipeline specific dashboard (after this batch of changes) and provide a link to it for more details. Wdyt? This dashboard is supposed to be just for SLOs and a new dashboard would be easier to modify if only your team will touch it.

dashboards/grafana-dashboard-rhtap-slos.configmap.yaml
```diff
@@ -1108,12 +909,12 @@ data:
                     "type": "prometheus",
                     "uid": "${datasource}"
                 },
-                "description": "The number of times the pipelines controller has restarted",
+                "description": "The rate that any of the pipeline controllers  have restarted",
```
amisstea, line 912:
```suggestion
                "description": "The rate that any of the pipeline controllers have restarted",
```

dashboards/grafana-dashboard-rhtap-slos.configmap.yaml
```diff
@@ -1122,7 +923,7 @@ data:
                         "showLineNumbers": false,
                         "showMiniMap": false
                     },
-                    "content": "<h1><center>The number of times the pipelines controller has restarted</center></h1>",
+                    "content": "<h1><center>The rate that any of the pipeline controllers  have restarted</center></h1>",
```
amisstea, line 926:
```suggestion
                    "content": "<h1><center>The rate that any of the pipeline controllers have restarted</center></h1>",
```

dashboards/grafana-dashboard-rhtap-slos.configmap.yaml
gabemontero: yeah I'm going to change it to `(alert remove, panel remove occurring soon)` and then will craft the PR to remove after this merges

but if you come up with something else before I push, I'll try to catch it and adjust @amisstea

gabemontero:
update pushed @amisstea PTAL thanks

amisstea (approved):

PR #363, opened 2024-09-02
feat(RELEASE-1031): define alerts for Release availability

theflockers:
this commit adds new alerts for a new Release Service SLO namely ReleaseServiceControllerManagerAvailability.

### DIFF
--- a/rhobs/alerting/data_plane/prometheus.release_service_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.release_service_alerts.yaml
@@ -59,3 +59,27 @@ spec:
         alert_team_handle: <!subteam^S03SVBS426R>
         runbook_url: https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/main/release-service/release-processing-latency.md?ref_type=heads
         team: release
+
+    - alert: ReleaseServiceControllerManagerAvailability
+      expr: |
+        (
+          avg_over_time
+          (
+            (sum(kube_pod_container_status_ready{namespace="release-service", container="manager"}))
+            [1h:]
+          )
+          /
+          max_over_time(sum(kube_pod_container_status_ready{namespace="release-service", container="manager"}) [1d:])
+        ) < 0.99
+      for: 15m
+      labels:
+        severity: critical
+        slo: "true"
+      annotations:
+        summary: >-
+          The Release Service Controller Manager must be available 99% of the time
+        description: >-
+          The Release Service Controller Manager is failing to be available for 99% of the time
+        alert_team_handle: <!subteam^S03SVBS426R>
+        runbook_url: https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/main/release-service/release-controller-availability.md?ref_type=heads
+        team: release
--- a/test/promql/tests/data_plane/release_service_test.yaml
+++ b/test/promql/tests/data_plane/release_service_test.yaml
@@ -4,7 +4,8 @@ rule_files:
 evaluation_interval: 1m
 
 tests:
-  - interval: 1m
+  - name: ReleaseServiceAlertOnPreProcessingSLOFailure
+    interval: 1m
     input_series:
       - series: 'release_pre_processing_duration_seconds_bucket{le="10", job="release", namespace="foo", source_cluster="cluster01"}'
         values: '1+8x59'
@@ -31,7 +32,8 @@ tests:
               runbook_url: https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/main/release-service/release-processing-latency.md?ref_type=heads
               team: release
 
-  - interval: 1m
+  - name: ReleaseServiceAlertOnValidationSLOFailure
+    interval: 1m
     input_series:
       - series: 'release_validation_duration_seconds_bucket{le="5", job="release", namespace="foo", source_cluster="cluster01"}'
         values: '1+8x59'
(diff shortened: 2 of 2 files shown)


yftacherzog (changes requested):
Can you add unit tests for the added alert?

theflockers:
> Can you add unit tests for the added alert?

Definitely. I totally forgot about that.

theflockers:
Also this new alert requires the `kube_pod_container_status_ready`, which will be added by the PR https://github.com/redhat-appstudio/infra-deployments/pull/4437/files

rhobs/alerting/data_plane/prometheus.release_service_alerts.yaml
```diff
@@ -59,3 +59,21 @@ spec:
         alert_team_handle: <!subteam^S03SVBS426R>
         runbook_url: https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/main/release-service/release-processing-latency.md?ref_type=heads
         team: release
+
+    - alert: ReleaseServiceControllerManagerAvailability
+      expr: |
+        rate(
+            sum( (sum(kube_pod_container_status_ready{namespace="release-service", container="manager"}) by(job) OR on() vector(0))) [10m:]
```
yftacherzog, line 66: What's the reason for having `sum` inside `sum` here?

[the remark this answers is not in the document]
theflockers: the comments a related. This metric give 1 or absent in case of pod not ready. To calculate the rate, I need it to be when absent, so the `vector` needs to be added. But this creates an extra item without labels, for instance:

when the data exists (or with 1)
```
kube_pod_container_status_ready{namespace="release-service"}  1
{}          0
```

About the corner case, that's the one explained  - if no data (which is the case when the pod is down) it gives 0.

I will look up how to emulate this in the test.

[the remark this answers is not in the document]
theflockers: I am trying to simplify the query using other metrics now that they are available.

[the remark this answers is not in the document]
theflockers: I fully changed the query to be more intuitive.

rhobs/alerting/data_plane/prometheus.release_service_alerts.yaml
```diff
@@ -59,3 +59,27 @@ spec:
...
         runbook_url: https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/main/release-service/release-processing-latency.md?ref_type=heads
         team: release
+
+    - alert: ReleaseServiceControllerManagerAvailability
+      expr: |
+        (
+          avg_over_time
+          (
+            (sum(kube_pod_container_status_ready{namespace="release-service", container="manager"}) by(namespace))
+            [1h:]
```
yftacherzog, line 69: Is it intentional that we have 1h at the numerator and 1d below?

test/promql/tests/data_plane/release_service_test.yaml
```diff
@@ -57,3 +57,24 @@ tests:
               alert_team_handle: <!subteam^S03SVBS426R>
               runbook_url: https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/main/release-service/release-validation-latency.md?ref_type=heads
               team: release
+
+  - interval: 1m
+    input_series:
+      - series: 'kube_pod_container_status_ready{namespace="release-service", container="manager"}'
```
yftacherzog, line 63: We have `sum` on the query, but we only have a single data series on the tests, so we don't test summing.

rhobs/alerting/data_plane/prometheus.release_service_alerts.yaml
```diff
@@ -59,3 +59,27 @@ spec:
...
+          )
+          /
+          max_over_time(sum(kube_pod_container_status_ready{namespace="release-service", container="manager"}) by(namespace)[1d:])
```
yftacherzog, line 72: If we only keep a single namespace (`namespace="release-service"`), then I think we don't need to sum by namespace, as all data series will have the same namespace anyhow. i.e. I think we can drop the `by(namespace)`. No?

test/promql/tests/data_plane/release_service_test.yaml
```diff
@@ -57,3 +57,24 @@ tests:
               alert_team_handle: <!subteam^S03SVBS426R>
               runbook_url: https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/main/release-service/release-validation-latency.md?ref_type=heads
               team: release
+
+  - interval: 1m
```
yftacherzog, line 61: I suggest adding a test case that will include data series from multiple clusters (i.e. entries with `source_cluster` with varying values). This way the tests can verify that we aggregate what we wanted to aggregate. E.g. you might not want to aggregate time series between different clusters.

rhobs/alerting/data_plane/prometheus.release_service_alerts.yaml
theflockers: Yes. The reason is that we might one of the metrics absent for a moment (maybe in the hour it checked it and I the calculation would be done with a missing node. Using one day (theoretically) I can have a  "good enough" assumption of how many nodes I should have up in that hour. I am open to suggestion.

test/promql/tests/data_plane/release_service_test.yaml
theflockers: Indeed. It makes  total sense.

test/promql/tests/data_plane/release_service_test.yaml
theflockers: indeed.

rhobs/alerting/data_plane/prometheus.release_service_alerts.yaml
theflockers: I will check. I tried without it, for something broke. I might have mixed problems when I tried it.

test/promql/tests/data_plane/release_service_test.yaml
theflockers: done.

test/promql/tests/data_plane/release_service_test.yaml
theflockers: added samples from different clusters.

rhobs/alerting/data_plane/prometheus.release_service_alerts.yaml
theflockers: removed the `by(..)`

rhobs/alerting/data_plane/prometheus.release_service_alerts.yaml
yftacherzog: I think we should have a test case for that scenario.

test/promql/tests/data_plane/release_service_test.yaml
yftacherzog: I think this is currently summing across all clusters. Is that the intention?

theflockers: yes. We consider the service down if the pod is down in any cluster.

rhobs/alerting/data_plane/prometheus.release_service_alerts.yaml
theflockers: added.

yftacherzog: Cool. Can you also add a test case in which the alert is not firing?

theflockers: added.

yftacherzog (approved):
lgtm

yftacherzog (changes requested):
@theflockers Can you squash the commits?

theflockers:
> @theflockers Can you squash the commits?

definitely.

yftacherzog (approved):
lgtm

PR #386, opened 2024-09-19
feat: adds recording rules to get tekton_pipelines metrics

theflockers:
adds recording rules to get tekton_pipelines metrics


theflockers:
does not apply.

PR #388, opened 2024-09-20
feat(KONFLUX-4508): Adding rules to detect Pipelines controller leases not distributed correctly

jhutar:


rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
```diff
@@ -183,3 +183,45 @@ spec:
             team: pipelines
             # link will be available once https://issues.redhat.com/browse/SRVKP-5899 is complete
             runbook_url: https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/main/pipeline-service/slos/pac-performance.md
+        - alert: PipelinesControllerPipelineRunRecLeasesNotDistributed
+          expr: count(count by (lease_holder) (kube_lease_owner{lease=~"tekton-pipelines-controller.github.com.tektoncd.pipeline.pkg.reconciler.pipelinerun.reconciler..*", lease_holder!=""})) <= 1
```
yftacherzog, line 187: Metric `kube_lease_owner` is not currently exported out of the cluster. Before it can be used, it needs to be exported. The instructions are here: https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/main/o11y/monitoring/tshoot-missing-metrics.md?ref_type=heads#missing-metric

Also, do you know how much does the identity of individual data series for this metric change over time? We have quota limits on the number of time series we hold in RHOBS, so we need to see that we're not exporting metrics for which new data series are added frequently based on the service workload.
Can you tell what's causing new data series to appear for that metric?

cc @amisstea @mftb

yftacherzog (changes requested):
@jhutar can you add unit tests for the new rules?

rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
jhutar: Thank you Yftach! Created https://github.com/redhat-appstudio/infra-deployments/pull/4583. Hope it is OK I also added to `LabelKeep` regexp.

Regarding how many labels combinations this mean, on `stone-stg-rh01` during last week I see 51. There are actually only 2 labels that change over time:

* `lease` - if we do not reconfigure Tekton Pipelines Controller to use different buckets count, there will be just 4 of these for every reconciler (so 12)
* `lease_holder` - this is a Pipelines Controller pod identification, so every time that pod restarts or so, new label value appears here (these pods are supposed to be stable, always running and restarts will be matter of upgrades or signs of cluster issues)

jhutar:
> @jhutar can you add unit tests for the new rules?

Sure, happy to do so. Could you please point me to some test as an example? I have no idea on where do they live.

amisstea:
@jhutar the data plane alerting rule unit tests are [here](https://github.com/redhat-appstudio/o11y/tree/main/test/promql/tests/data_plane). Refer to any of the existing tests for examples.

jhutar:
Hello @amisstea and @yftacherzog . I have added tests and when thinking about these also added few more alerts to cover when the metric (or metric with expected labels) is missing - it can happen, but only for something like 2 seconds when leases are being reconciled.

Please take a look and let me know if it looks good.

jhutar:
Fixed git-lint and yaml-lint issues as it is possible to install these locally, but I do not understand what is the problem with the tests, as they are passing for me locally and I do not have access to PR check logs. Could you please help here if it is still failing now?

rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
```diff
@@ -183,3 +183,75 @@ spec:
             team: pipelines
             # link will be available once https://issues.redhat.com/browse/SRVKP-5899 is complete
             runbook_url: https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/main/pipeline-service/slos/pac-performance.md
+        - alert: PipelinesControllerPipelineRunRecLeasesNotDistributed
+          expr: count by (source_cluster) (count by (lease_holder, source_cluster) (kube_lease_owner{lease=~"tekton-pipelines-controller.github.com.tektoncd.pipeline.pkg.reconciler.pipelinerun.reconciler..*", lease_holder!=""})) <= 1
```
yftacherzog, line 187: I think we should have test cases with more than one `source_cluster`, especially if we aggregate by this label.

yftacherzog: Operators precedence here is a little challenging. What comes first? the `<=` or the count? Maybe we can add parentheses to make the statement clearer.

rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
```diff
@@ -183,3 +183,75 @@ spec:
...
+          labels:
+            severity: warning
+          annotations:
+            summary: Tekton Pipelines controller PipelineRun reconciler leases metric is missing
+            description: Tekton Pipelines controller PipelineRun reconciler leases metric is missing or lease_holder is empty.
+            alert_team_handle: <!subteam^S04PYECHCCU>
+            team: pipelines
+            runbook_url: https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/main/o11y/alert-rule-PipelinesControllerLeasesNotDistributed.md
+        - alert: PipelinesControllerTaskRunRecLeasesMetricMissing
+          expr: absent(kube_lease_owner{lease=~"tekton-pipelines-controller.github.com.tektoncd.pipeline.pkg.reconciler.taskrun.reconciler..*", lease_holder!=""}) == 1
```
yftacherzog, line 237: Does this value of `lease` only exist in one cluster? Suppose that it exists in more than one, this alert will fire only when it's missing from both. No? I suggest adding a test case with more than a singled `source_cluster` for this alert as well.

jhutar: Hello!

These leases are there in all the clusters, so you are right! If I add `absent(sum by (source_cluster) (...)) == 1` would it help?

Yes, will add tests with multiple `source_cluster`s.

jhutar: Hello @yftacherzog . I was thinking about it and what would you think about changing the rule to:
```
(count by (source_cluster) (kube_lease_owner{lease=~"tekton-pipelines-controller.github.com.tektoncd.pipeline.pkg.reconciler.pipelinerun.reconciler..*", lease_holder!=""}) or on() vector(0)) == 0
```
?

rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
jhutar: IMO this thread is resolved now.

rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
yftacherzog: We saw many examples using `or on() vector(0)`, but when trying to use it, we could not find a single scenario in which that worked for us.
Having said that, it might work... The important bit is to create the relevant test case that represents what you're trying to solve and see that it indeed solves it.

it might be that incorporating the [absent](https://prometheus.io/docs/prometheus/latest/querying/functions/#absent) query function into this rule (or another complementing rule) can simplify the whole thing.

jhutar: Thank you. I was not able to figure out how to use `absent(...)` to check metric on cluster A is missing while it is still there on cluster B. Will take a look into it again.

yftacherzog: What is it that you're trying to alert on in this case?

jhutar: I want to ensure that lease_holder is always set:
```
kube_lease_owner{lease=~"tekton-pipelines-controller.github.com.tektoncd.pipeline.pkg.reconciler.pipelinerun.reconciler..*", lease_holder!=""}
```
Basically that this ^^^ is not absent.

yftacherzog: It might be that [absent_over_time](https://prometheus.io/docs/prometheus/latest/querying/functions/#absent_over_time) is a better fit here.

The reason being that the metric only gives a value of 1 when the state it represents happens, and stops providing values when it's not the case anymore, so I believe that as long as the data series exists, it will not be absent.

yftacherzog: In any case, you might benefit from dealing with the different alerts you're working on here on separate PRs. If they don't strictly depend on each other then it might be that you can have some of them merged while figuring out the others.

jhutar: Hello! Exactly. Just submitted simplified version. I gave up on the "metric is absent" bit of the alert and test (as that actually should not happen and if it does, it probably means whole controller is down), and only stick to "lease_holder is not empty". It is far easier to create rule and test for it now. Please take a look.

rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
```diff
@@ -183,3 +183,75 @@ spec:
...
+          annotations:
+            summary: Tekton Pipelines controller ResolutionRequest reconciler leases not distributed across more than one pod
+            description: >-
+              Tekton Pipelines controller ResolutionRequest reconciler leases not distributed across more than one pod
+              on cluster {{ $labels.source_cluster }}.
+            alert_team_handle: <!subteam^S04PYECHCCU>
+            team: pipelines
+            runbook_url: https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/main/o11y/alert-rule-PipelinesControllerLeasesNotDistributed.md
+        - alert: PipelinesControllerPipelineRunRecLeasesHolderMissing
+          expr: count by (source_cluster) (kube_lease_owner{lease=~"tekton-pipelines-controller.github.com.tektoncd.pipeline.pkg.reconciler.pipelinerun.reconciler..*", lease_holder=""}) > 0
```
yftacherzog, line 226: With this expression, and because that metric will either give `1` or `_` (nothing), I wonder what will happen if you'd have a test case in which you had some `1`s and then had nothing. I'm not sure it won't alert.
i.e. have a test case in which you have a value of `1` for 6 minutes and then have `_` for about 10 minutes. When will the alarm stop firing?

jhutar: Added tests for that: One with `_x6 1x14` and another one with `1x6 _x14` - no alert in both cases.

Second case is obviously not a healthy state, but as I said above, I decided not try to detect that state by these rules. Hope it makes sense.

yftacherzog: > I decided not try to detect that state by these rules.
I understand what you mean now.

If that is the case, wouldn't be enough to have a simpler expressions such as this one:
```
kube_lease_owner{lease=~"tekton-pipelines-controller.github.com.tektoncd.pipeline.pkg.reconciler.pipelinerun.reconciler..*", lease_holder=""}
```

jhutar: Hmm, did not knew it is possible, but with that change, I'm getting so I guess I would need some aggregation anyway?
```
  FAILED:
    alertname: PipelinesControllerPipelineRunRecLeasesHolderMissing, time: 10m, 
        exp:[
            0:
              Labels:{alertname="PipelinesControllerPipelineRunRecLeasesHolderMissing", severity="warning", source_cluster="cluster01"}
              Annotations:{alert_team_handle="<!subteam^S04PYECHCCU>", description="Tekton Pipelines controller PipelineRun reconciler leases lease_holder is empty on cluster cluster01.", runbook_url="https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/main/o11y/alert-rule-PipelinesControllerLeasesNotDistributed.md", summary="Tekton Pipelines controller PipelineRun reconciler leases holder is missing", team="pipelines"}
            ], 
        got:[
            0:
              Labels:{alertname="PipelinesControllerPipelineRunRecLeasesHolderMissing", lease="tekton-pipelines-controller.github.com.tektoncd.pipeline.pkg.reconciler.pipelinerun.reconciler.00-of-04", namespace="openshift-pipelines", severity="warning", source_cluster="cluster01"}
              Annotations:{alert_team_handle="<!subteam^S04PYECHCCU>", description="Tekton Pipelines controller PipelineRun reconciler leases lease_holder is empty on cluster cluster01.", runbook_url="https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/main/o11y/alert-rule-PipelinesControllerLeasesNotDistributed.md", summary="Tekton Pipelines controller PipelineRun reconciler leases holder is missing", team="pipelines"},
            1:
              Labels:{alertname="PipelinesControllerPipelineRunRecLeasesHolderMissing", lease="tekton-pipelines-controller.github.com.tektoncd.pipeline.pkg.reconciler.pipelinerun.reconciler.01-of-04", namespace="openshift-pipelines", severity="warning", source_cluster="cluster01"}
              Annotations:{alert_team_handle="<!subteam^S04PYECHCCU>", description="Tekton Pipelines controller PipelineRun reconciler leases lease_holder is empty on cluster cluster01.", runbook_url="https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/main/o11y/alert-rule-PipelinesControllerLeasesNotDistributed.md", summary="Tekton Pipelines controller PipelineRun reconciler leases holder is missing", team="pipelines"},
            2:
              Labels:{alertname="PipelinesControllerPipelineRunRecLeasesHolderMissing", lease="tekton-pipelines-controller.github.com.tektoncd.pipeline.pkg.reconciler.pipelinerun.reconciler.02-of-04", namespace="openshift-pipelines", severity="warning", source_cluster="cluster01"}
              Annotations:{alert_team_handle="<!subteam^S04PYECHCCU>", description="Tekton Pipelines controller PipelineRun reconciler leases lease_holder is empty on cluster cluster01.", runbook_url="https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/main/o11y/alert-rule-PipelinesControllerLeasesNotDistributed.md", summary="Tekton Pipelines controller PipelineRun reconciler leases holder is missing", team="pipelines"},
            3:
              Labels:{alertname="PipelinesControllerPipelineRunRecLeasesHolderMissing", lease="tekton-pipelines-controller.github.com.tektoncd.pipeline.pkg.reconciler.pipelinerun.reconciler.03-of-04", namespace="openshift-pipelines", severity="warning", source_cluster="cluster01"}
              Annotations:{alert_team_handle="<!subteam^S04PYECHCCU>", description="Tekton Pipelines controller PipelineRun reconciler leases lease_holder is empty on cluster cluster01.", runbook_url="https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/main/o11y/alert-rule-PipelinesControllerLeasesNotDistributed.md", summary="Tekton Pipelines controller PipelineRun reconciler leases holder is missing", team="pipelines"}
            ]
```

yftacherzog: It depends on what you want to see. If you don't aggregate, it gives you more details on the lease that's missing a holder. Would that be valuable for whoever handles the alert?

If you prefer it aggregated then count by cluster, as you did, sounds good.

In any case, if you don't aggregate, I suppose you can have fewer data series for the test.

jhutar:
Awesome, thank you. Squashed.

yftacherzog:
> Awesome, thank you. Squashed.

I think something went wrong there

jhutar:
> > Awesome, thank you. Squashed.
> 
> I think something went wrong there

Fixed commit message.

yftacherzog (approved):
lgtm

PR #390, opened 2024-09-23
chore: update konflux_up alert routing

yftacherzog:
The konflux_up metric had a routing key which did not correspond to any routing rule in app-interface. Here we change it to a value that corresponds to an existing routing statement.


yftacherzog:
I see that this metric is not being routed correctly as its routing key did not exist in the routing rules in app-interface. I think we should use one of the routing keys that already exist instead of creating a new one in app-interface.

hmariset (approved):
/lgtm

mftb (approved):
/lgtm
/approve

amisstea (approved):

PR #393, opened 2024-09-25
chore: delete alert for grafana konflux-up

yftacherzog:
We have a generic check for absent konflux_up metrics so we don't need to maintain a specific alert for the grafana datasource konflux_up.


hmariset (approved):
/lgtm

PR #394, opened 2024-09-26
fix: fixing availability query

theflockers:
this commit fixes the query that fetches the availability data, by ignoring duplicate data
and not counting data with value 0.


theflockers:
changed back to draft as I found a better solution for the query

theflockers:
> changed back to draft as I found a better solution for the query

leaving the `1h` change for now as the promtool fails with the new query and the alert needs to be fixed for now.

mftb (approved):

PR #398, opened 2024-10-02
fix: SLO alert improvement

theflockers:
this commit decreases the observation period and
moves the aggregation function back to sum.

testing historical data for the observation, this
query seems to be the most accurate so far conside-
ring the eventual occurence of the service in clusters not releasing.


mftb (approved):
/lgtm
/approve

PR #408, opened 2024-11-11
fix: release SLO alert should consider clusters in use

theflockers:
this PR change the release's controller SLO alert to consider only the clusters used for releasing software, namely stone-prd-rh01 and stone-prod-p02.


ralphbean:
If you rebase on `main` that policy failure should go away.

pacho-rh (approved):

PR #411, opened 2024-11-12
Add a crashloopbackoff alert for tekton

ralphbean:
I noticed that the PipelinePodsRepeatedRestarts alert has a problem: at first, when pods are restarting, it fires. Good! But then, after a short while, the pod enters CrashLoopBackOff and it starts restarting less frequently. This causes PipelinePodsRepeatedRestarts to stop firing, making it appear as if the outage has resolved itself, when in fact it hasn't.

Adding this second alert will cause a second alert to fire when the situation cements itself into a CrashLoopBackOff.

Perhaps we should just remove the PipelinePodsRepeatedRestarts alert and replace it with this second one? Or - we could try both for a while and see how they play together before deciding?

### DIFF
--- a/rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
@@ -25,6 +25,22 @@ spec:
             alert_team_handle: <!subteam^S04PYECHCCU>
             team: pipelines
             runbook_url: https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/main/pipeline-service/sre/core-pipeline-controller-repeated-restarts.md
+        - alert: PipelinePodsCrashLoopBackOff
+          expr: |
+            sum by (source_cluster) (max_over_time(kube_pod_container_status_waiting_reason{namespace="openshift-pipelines", reason="CrashLoopBackOff"}[3m]) or vector(0)) > 0
+          for: 3m
+          labels:
+            severity: critical
+            slo: "true"
+          annotations:
+            summary: >-
+              Tekton controller is in a crashloop
+            description: >-
+              Tekton controllers on cluster {{ $labels.source_cluster }} has degraded into CrashLoopBackOff status and is not starting up.
+            alert_team_handle: <!subteam^S04PYECHCCU>
+            team: pipelines
+            # Same runbook as the repeated restarts alert
+            runbook_url: https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/main/pipeline-service/sre/core-pipeline-controller-repeated-restarts.md
         - alert: PipelineControllerDeadlock
           expr: |
             sum by (source_cluster) (increase(pipelinerun_kickoff_not_attempted_count[2m])) > 0


mftb:
@ralphbean there seems to be a `gitlint` issue

```
+ gitlint --fail-without-commits --commits origin/main..HEAD
WARNING: I4 - ignore-by-author-name: gitlint will be switching from using Python regex 'match' (match beginning) to 'search' (match anywhere) semantics. Please review your ignore-by-author-name.regex option accordingly. To remove this warning, set general.regex-style-search=True. More details: https://jorisroovers.github.io/gitlint/configuration/#regex-style-search
1: CT1 Title does not follow ConventionalCommits.org format 'type(optional-scope): description': "Add a crashloopbackoff alert for tekton"
```

mftb:
Other than that, LGTM

mftb (approved):
/lgtm
/approve

ralphbean:
Adding `rpms-signature-scan` task in https://github.com/redhat-appstudio/o11y/pull/413 to address that policy failure.

arewm (approved):
thanks for this!

mftb:
@ralphbean looks like the enterprise contract is failing, I've looked into it and can't really figure out why. Do you have any insight here?

rhobs/alerting/data_plane/prometheus.pipeline_alerts.yaml
```diff
@@ -25,6 +25,22 @@ spec:
             alert_team_handle: <!subteam^S04PYECHCCU>
             team: pipelines
             runbook_url: https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/main/pipeline-service/sre/core-pipeline-controller-repeated-restarts.md
+        - alert: PipelinePodsCrashLoopBackOff
+          expr: |
+            sum by (source_cluster) (max_over_time(kube_pod_container_status_waiting_reason{namespace="openshift-pipelines", reason="CrashLoopBackOff"}[3m]) or vector(0)) > 0
```
zanssa, line 30: Could we please add one more filter here to only alert when the main pods are not in a good status? The namespace has other pods and based on the comments [here](https://issues.redhat.com/browse/KONFLUX-3581?focusedId=25192463&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-25192463) we are interested in the controllers started with tekton or pipeline as code. Please let me know if this does not apply anymore.

zanssa:
>> Perhaps we should just remove the PipelinePodsRepeatedRestarts alert and replace it with this second one? Or - we could try both for a while and see how they play together before deciding? 


hmm. The idea of using `kube_pod_container_status_restarts_total` metric is to alert when there is a restart without specifying the reason. The new alert would only alert if there is a crashloop. I don't think there would be alert when there is a restart due to resource limitations.

Are we planning to bump the grafana dashboard to include this alert? I don't see there is an update in app-interface or open MR for that.

zanssa:
Does it make sense to have tests for the alert? something similar to these: https://github.com/redhat-appstudio/o11y/tree/main/test/promql/tests/data_plane

PR #353, opened 2024-08-27
add build team availability alerts

tnevrlka:
Jira: [STONEBLD-2651](https://issues.redhat.com/browse/STONEBLD-2651)

Add availability metrics for the build team which fire after five minutes of not having the `konflux_up` parameter
- Add GitHubAppFailureAlert for build-service
- Add QuayFailureAlert for image-controller

The alerts include `runbook_url` to currently non-existing SOPs. There is a [MR open which adds the SOPs](https://gitlab.cee.redhat.com/konflux/docs/sop/-/merge_requests/179), but it's not yet merged (will mark this PR ready for review after it's merged)

PR #423, opened 2024-11-26
feat: add availability alerts for the build team

tnevrlka:
- Add GitHubAppFailureAlert which should trigger when GitHub App is not working in build-service
- Add QuayFailureAlert which should trigger when Quay is not working in image-controller
- Add tests for both of the alerts which test that the alerts are not triggered on different checks and services than expected

Similar to https://github.com/redhat-appstudio/o11y/pull/353 which had to be reverted because the alerts were triggering in unexpected services and checks. 
- `(konflux_up offset 1h)` has been changed to  `konflux_up{service="build-service", check="github"} offset 1h)` (similarly for QuayFailureAlert) in the `expr` field which should result in the alerts triggering only on the correct combination of `service` and `check`
- `service` and `check` have been switched in the old PR, that is now fixed
- More tests have been added to test that the alerts are not triggered on different checks and services than expected

### DIFF
--- a/rhobs/alerting/data_plane/prometheus.build_service_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.build_service_alerts.yaml
@@ -1,0 +1,23 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: rhtap-build-service-github-app-alerting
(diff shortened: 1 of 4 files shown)


mftb (approved):
/lgtm
/approve

PR #424, opened 2024-11-28
fix: fixing the expression for the 'Degraded service instances' widget

mftb:


dashboards/grafana-dashboard-konflux-availability-exporter.configmap.yaml
```diff
@@ -283,7 +283,7 @@ data:
               },
               "editorMode": "code",
               "exemplar": false,
-              "expr": "sum( sum by (service, check) (konflux_up{service=~\"$service\"} == 0))",
+              "expr": "sum( count by (service, check) (konflux_up{service=~\"$service\"} == 0))",
```
theflockers, line 286: The only thing to notice is, if the number displayed represents the services degraded per cluster, so this one is correct. IE 1 service degrade in 3 clusters, so it should display 3

If the desired is should the number of services degraded, not taking in consideration the cluster, both should use `count` in both. IE:  1 service degrade all around, the it should display 1.

pacho-rh: Talking with @mftb, the correct case is what he intended with this change (number of services degraded at per-cluster granularity)

pacho-rh (approved):

PR #469, opened 2025-02-18
feat(KFLUXVNGD-197): Add HypershiftAWSQuotaAlmostFull alerting rule

amisstea:
Depends on:
- https://github.com/redhat-appstudio/infra-deployments/pull/5596
- https://gitlab.cee.redhat.com/konflux/docs/sop/-/merge_requests/265

PR #471, opened 2025-02-20
Revert "feat(KFLUXVNGD-197): Add HypershiftAWSQuotaAlmostFull alerting rule"

amisstea:
Reverts redhat-appstudio/o11y#469

I discovered the quota isn't being enforced as I expected (it's actually namespaced, not cluster-scoped). Reverting for now.


amisstea:
On second thought, this can still be enforced with a ClusterResourceQuota so I'll make those changes instead.

amisstea:
According to the OCP docs, `ClusterResourceQuotas` must define a `selector` that should match active namespaces to the order of dozens at most. The namespaces will contend on this quota during object creation so it can create a performance bottleneck. We must plan to scale into the hundreds of active namespaces at a minimum so this might cause problems for us.

With no great option for enforcing a global quota I'm back to the point where reverting this alerting rule is probably the best option. @mftb @mike-kingsbury please review.

mike-kingsbury (approved):

PR #474, opened 2025-02-24
fix: unblock mintmaker update to go-toolset due to change in dir perms

mike-kingsbury:
The go-toolset builder image changed the directory perms of /usr/bin in 1.22 to remove the group writable bit.  This causes the mintmaker proposed changes to fail as go can not compile the 'exporters' binary into /bin (symlinked to /usr/bin).  This change writes the exporter binary to /tmp during compilation and then copies it to the image.


raks-tt (approved):

mftb (approved):
/lgtm
/approve

PR #485, opened 2025-03-14
chore: ignore PodNotReady alerts in mintmaker namespace

qixiang:
mintmaker runs pipelineruns for users within its own namespace, these pipelineruns are renovate jobs, we don't need to be alerted about such errors.

### DIFF
--- a/rhobs/alerting/data_plane/prometheus.pod_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.pod_alerts.yaml
@@ -36,7 +36,7 @@ spec:
         runbook_url: https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/main/o11y/alert-rule-crashLoopBackOff.md
     - alert: PodNotReady
       expr: |
-            kube_pod_status_phase{phase=~"Pending|Unknown|Failed", namespace!~"(.*-tenant|.*-env|openshift-.*|kube-.*|default|tekton-ci|build-templates-e2e|konflux-ci)"} == 1
+            kube_pod_status_phase{phase=~"Pending|Unknown|Failed", namespace!~"(.*-tenant|.*-env|openshift-.*|kube-.*|default|tekton-ci|build-templates-e2e|konflux-ci|mintmaker)"} == 1
             unless ignoring (phase) (kube_pod_status_unschedulable == 1)
       for: 30m
       labels:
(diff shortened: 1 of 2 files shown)


staticf0x (approved):
Good idea in my opinion

kubasikus:
This will block any alerts from the mintmaker namespace - including alerts for mintmaker-controller. Do you have another alert in place for the controller?

Also,
The CI pipeline seems to be failing due to commit not being signed:

`1: CC1 Body does not contain a 'Signed-off-by' line`

qixiang:
> This will block all pod alerts from the mintmaker namespace - including alerts for mintmaker-controller. Do you have another alert in place for the controller?

We don't have alert for the controller yet, but we have one epic planned for next quarter to implement SLOs for mintmaker. Not sure if this is acceptable to, if not, we'll try to add alert for controller first.

> Also, The CI pipeline seems to be failing due to commit not being signed:
> 
> `1: CC1 Body does not contain a 'Signed-off-by' line`

kubasikus:
> We don't have alert for the controller yet, but we have one epic planned for next quarter to implement SLOs for mintmaker. Not sure if this is acceptable, if not, we'll try to add alert for controller first.

It depends - what happens if the controller goes down? How will it affect users?

qixiang:
> > We don't have alert for the controller yet, but we have one epic planned for next quarter to implement SLOs for mintmaker. Not sure if this is acceptable, if not, we'll try to add alert for controller first.
> 
> It depends - what happens if the controller goes down? How will it affect users?

If the controller goes down, users may not receive dependency update PRs in their repositories. However, this isn't very critical - even if it's down for several tens of minutes or a few hours, it won't cause significant impact. Once the controller recovers, users will receive their update PRs again.

staticf0x:
We also have additional custom monitoring dashboard in Splunk that can be used in the meantime.

kubasikus:
Ack, then I agree that we can disable this for the time being.

One more thing - You should also add a test for this in the unit test file https://github.com/redhat-appstudio/o11y/blob/487839fd1b6f19097c307e1b8aefb70104b1d5f6/test/promql/tests/data_plane/unschedulable_pods_test.yaml#L47 .

Once that's done, I'll get this approved.

kubasikus:
@martysp21 @mike-kingsbury Can you give this a quick look and approve? :)

martysp21 (approved):
With the Test in place LGTM

qixiang:
PR is updated with new test case.

qixiang:
@kubasikus @mike-kingsbury @martysp21 Could you please review the updated PR and merge it if everything looks good? We've been receiving numerous PodNotReady alerts from the mintmaker namespace, and this PR aims to suppress these alerts.

PR #522, opened 2025-04-23
Pod exceeding cpu & memory requested alerts

FaisalAl-Rayes:
[PVO11Y-4645](https://issues.redhat.com/browse/PVO11Y-4645)

This is to warn of pods running above what they requested which puts them at the risk of being evicted in the case of cluster pressure, not only that but it also should signal a change in the behaviour of the pod and might require a revision of the requested resources.


mike-kingsbury:
I'm a little concerned how noisy this will be.  I'd be curious to know how many pods operate under their requested amount of mem & cpu.  I know we can always disable this alert if it does become an issue.  

It appears there's an issue with the CI pipeline, will approve once that's fixed.

FaisalAl-Rayes:
@mike-kingsbury

I would imaging it could be noisy but in the hopes that PRs will start coming in to set the appropriate requests and limits for the resources.

Also the CI pipeline failed on a yaml lint issue... retesting

mike-kingsbury (approved):
LGTM...

mike-kingsbury (approved):

FaisalAl-Rayes:
@raks-tt PTAL

missing one more review

PR #525, opened 2025-04-24
feat(KONFLUX-7900): Do not trigger alert for every pod

jhutar:
This rule was too noisy. This change makes it to fire only once when one of the ETCD pods (or all of them) cross the line.


mike-kingsbury (approved):
LGTM

raks-tt (approved):
/lgtm
/ok-to-test

PR #532, opened 2025-04-29
feat(PVO11Y-4703): implement cluster filtering

martysp21:


dashboards/grafana-dashboard-konflux-availability-exporter.configmap.yaml
```diff
@@ -591,7 +591,7 @@ data:
         "list": [
           {
             "current": {
-              "selected": true,
+              "selected": false,
```
TominoFTW, line 594: What does this change mean? Not it will be empty and user needs to select it by hand from dropdown?

martysp21: Even though it would default out to "All", the change is unnecessary, so I changed it back in all instances.

TominoFTW (approved):
LGTM!

raks-tt:
LGTM

raks-tt (approved):

PR #541, opened 2025-05-08
fix(PVO11Y-4656): Update insufficient members alert threshold

gcpsoares:
- Update insufficient members alert threshold

### DIFF
--- a/rhobs/alerting/data_plane/prometheus.cluster_capacity_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.cluster_capacity_alerts.yaml
@@ -87,7 +87,7 @@ spec:
 
       - alert: EtcdInsufficientMembers
         expr: |
-          (count(etcd_server_id) by (source_cluster) % 2) == 1
+          (count(etcd_server_id) by (source_cluster) % 2) == 0
         for: 1m
         labels:
           severity: warning
--- a/test/promql/tests/data_plane/cluster_capacity_test.yaml
+++ b/test/promql/tests/data_plane/cluster_capacity_test.yaml
@@ -165,12 +165,10 @@ tests:
 
   - interval: 1m
     input_series:
-      # Etcd No Leader failures within threshold limit, so it will be alerted.
+      # Etcd insufficient members threshold limit, so it will be alerted.
       - series: 'etcd_server_id{source_cluster="cluster01",instance="instance1"}'
         values: '1+0x60'
       - series: 'etcd_server_id{source_cluster="cluster01",instance="instance2"}'
-        values: '1+0x60'
-      - series: 'etcd_server_id{source_cluster="cluster01",instance="instance3"}'
         values: '1+0x60'
 
     alert_rule_test:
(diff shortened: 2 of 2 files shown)


rhobs/alerting/data_plane/prometheus.cluster_capacity_alerts.yaml
```diff
@@ -87,7 +87,7 @@ spec:
 
       - alert: EtcdInsufficientMembers
         expr: |
-          (count(etcd_server_id) by (source_cluster) % 2) == 1
+          (count(etcd_server_id) by (source_cluster) % 2) == 0
```
raks-tt, line 90: My only concern here, we need to make sure all three etcd_server_id are present. This alert currently won't trigger if only 1 shows up but this needs to be accounted to as well.

test/promql/tests/data_plane/cluster_capacity_test.yaml
```diff
@@ -194,6 +192,9 @@ tests:
         values: '1+0x60'
       - series: 'etcd_server_id{source_cluster="cluster01",instance="instance2"}'
         values: '1+0x60'
+      - series: 'etcd_server_id{source_cluster="cluster01",instance="instance3"}'
+        values: '1+0x60'
+
     alert_rule_test:
       - eval_time: 11m
         alertname: EtcdInsufficientMembers
```
raks-tt, line 200: Please add a new line here

raks-tt (changes requested):

test/promql/tests/data_plane/cluster_capacity_test.yaml
gcpsoares: New line included.

rhobs/alerting/data_plane/prometheus.cluster_capacity_alerts.yaml
gcpsoares: Other alert will cover 1 instance up situation.

raks-tt, line 90: If all etcd instances are down or not reporting metrics, this expression might not trigger as the metric itself might not be present, not just equal to 0.

So I am proposing something like:
`count(etcd_server_has_leader{namespace="openshift-etcd"} == 1) == 0`
which actually ensures that it is checking that there is a leader and then captures the count of leaders, so if no leader metric is present, the count will be zero, triggering the alert.

wdyt?

gcpsoares: ETCD has no leader alert will cover similar situation. Another alert will be created to count instances up.

test/promql/tests/data_plane/cluster_capacity_test.yaml
```diff
@@ -189,11 +187,15 @@ tests:
...
         values: '1+0x60'
+      - series: 'etcd_server_id{source_cluster="cluster01",instance="instance3"}'
+        values: '1+0x60'
+
     alert_rule_test:
       - eval_time: 11m
-        alertname: EtcdInsufficientMembers
\ No newline at end of file
+        alertname: EtcdInsufficientMembers
+
```
mftb, line 201: This extra blank line should be removed.

raks-tt: yes add a new line but not an extra blank space

gcpsoares: removed blank space

mftb (approved):
/lgtm
/approve

pacho-rh (approved):

raks-tt (approved):

PR #573, opened 2025-05-21
fix(KFLUXVNGD-319): make project-controller alarm less sensitive

yftacherzog:
The alert is firing every once in a while and till now was always resolving itself. Given this and the nature of the work done with this controller (very sporadic and not very time sensitive) we can make the alert less sensitive.


amisstea (approved):

avi-biton (approved):
/lgtm

PR #576, opened 2025-05-22
feat(PVO11Y-4779): Dynamic multi-exporter build and runtime selection

FaisalAl-Rayes:
Allow to build more than one exporter and select the one to use using the "cmd" of the image.

### DIFF
--- a/config/exporters/monitoring/grafana/base/prometheus-exporter-service.yaml
+++ b/config/exporters/monitoring/grafana/base/prometheus-exporter-service.yaml
@@ -116,6 +116,7 @@ spec:
           runAsNonRoot: true
       - name: exporters
         image: exporter:latest
+        args: ["dsexporter"]
         resources:
           limits:
             cpu: 100m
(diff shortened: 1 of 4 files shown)


config/exporters/monitoring/grafana/base/prometheus-exporter-service.yaml
```diff
@@ -116,6 +116,7 @@ spec:
           runAsNonRoot: true
       - name: exporters
         image: exporter:latest
+        args: ["dsexporter"]
```
kubasikus, line 119: Once this change is accepted, we should test it in staging cluster via https://github.com/redhat-appstudio/infra-deployments/blob/main/components/monitoring/grafana/staging/kustomization.yaml . Then roll it out to prod as well.

kubasikus (approved):
This seems to work well and sets us up for the follow up work of adding a new exporter in https://github.com/redhat-appstudio/o11y/pull/574

mike-kingsbury (approved):
LGTM

PR #581, opened 2025-05-26
fix(PVO11Y-4776): Update original cluster capacity UID reference

gcpsoares:
-Update UID to keep reference of original dashboard UID.


mftb (approved):
/lgtm
/approve
/ok-to-test

kubasikus (approved):

PR #587, opened 2025-06-05
feat(PVO11Y-4764): Add Kanary signal and error alerts with tests

FaisalAl-Rayes:


FaisalAl-Rayes:
@mftb The ideas was to first have it ping us o11y for a bit then we can make a PR for it to be an SLO alert.

if kanary_up is ever 0 it means that there were 3 consecutive failures.

We would need to create an slo document for that as well in [konflux/docs/sop/o11y](https://gitlab.cee.redhat.com/konflux/docs/sop/-/tree/main/o11y/monitoring?ref_type=heads)
according to [docs on adding an slo alert](https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/b4b6d67130ab3fc4d6a58efdea760da40938efea/o11y/monitoring/adding-alert.md)

rhobs/alerting/data_plane/prometheus.kanary_alerts.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: rhtap-kanary-alerting-rules
+  labels:
+    tenant: rhtap
+spec:
+  groups:
```
mftb, line 8: All alerts are missing the alert_routing_key or alert_team_handle. IMO you can use the combined Infra and SPRE one that we're using in the etcd alerts, check: https://github.com/redhat-appstudio/o11y/blob/main/rhobs/alerting/data_plane/prometheus.cluster_capacity_alerts.yaml

Basically, for the SLO alert you define a alert_team_handle and for the other alerts, a alert_routing_key

FaisalAl-Rayes: Though how about adding an alert_team_handle to just our team or even just users (Me and Jakub) for a bit to see the state of it before changing the alert_team_handle. What do you think?

mftb (approved):
/lgtm
/approve

rhobs/alerting/data_plane/prometheus.kanary_alerts.yaml
```diff
@@ -0,0 +1,58 @@
...
+      - alert: KanarySignalTriggered
+        expr: kanary_up == 0
+        for: 0m
+        labels:
+          severity: critical
+          slo: "true"
+        annotations:
+          alert_team_handle: '<@U03S3DRLPH6> <@U067S2N67N3>'
+          team: o11y
+          summary: "Kanary signal has been triggered for cluster {{ $labels.exported_instance }}"
```
raks-tt, line 21: any reason we are using `exported_instance` instead of `source_cluster`?

raks-tt:
do we an SOP for troubleshooting that we can link for these alerts?

rhobs/alerting/data_plane/prometheus.kanary_alerts.yaml
```diff
@@ -0,0 +1,58 @@
...
+  labels:
+    tenant: rhtap
+spec:
+  groups:
+  - name: kanary_alerts
+    interval: 1m
+    rules:
+      - alert: KanarySignalTriggered
+        expr: kanary_up == 0
+        for: 0m
```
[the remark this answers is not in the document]
FaisalAl-Rayes: @raks-tt The reasoning behind this is that in the exporter, it sets the kanary_up metric value to 0 only when we have 3 consecutive confirmed (by the result written to the database) failures in the load tests on a given cluster. Normally these tests are running 1 per hour. so if we ever get kanary_up=0 we need to fire the alert immediately.

rhobs/alerting/data_plane/prometheus.kanary_alerts.yaml
FaisalAl-Rayes: @raks-tt The reasoning behind this that:

- The `source_cluster` is just where the metric is coming **from**.
- The `exported_instance` is the tested cluster of which the kanary_up metric is addressing. In other words who is this metric **about**.

So in the alert summary we would be interested in which cluster failed the tests, not where the metic came from.

FaisalAl-Rayes:
@raks-tt regarding the SOP, As of now there is no document for that @kubasikus will discuss about this soon, to see if we can construct something that would be useful to know there.

At the top of my head I can think of having the links of where the pipelines are running to check out the latest ones and see where something failed.

kubasikus:
> At the top of my head I can think of having the links of where the pipelines are running to check out the latest ones and see where something failed.

Since the Alert will go out to me and Faisal in the next week or so, we can take the time to write up the SOP after this is merged. I've spun off a task for us here https://issues.redhat.com/browse/PVO11Y-4801 .

rhobs/alerting/data_plane/prometheus.kanary_alerts.yaml
kubasikus: Sure, let's add a custom label to the metrics to identify the cluster where the load test ran. I'm thinking "tested_cluster" or "kanary_cluster" might be a good label name.

FaisalAl-Rayes: Created a ticket for it here.
[PVO11Y-4802](https://issues.redhat.com/browse/PVO11Y-4802)

FaisalAl-Rayes:
updated the alerts and tests with changes related to [this discussion](https://github.com/redhat-appstudio/o11y/pull/587#discussion_r2135254213)

Issue: [PVO11Y-4802](https://issues.redhat.com/browse/PVO11Y-4802)
PRs: https://github.com/redhat-appstudio/o11y/pull/597 https://github.com/redhat-appstudio/infra-deployments/pull/6620

PR #589, opened 2025-06-06
chore: rename kanary exporter service and deployment names

FaisalAl-Rayes:
The reason behind this is to be able to distinguish between the dsexporter and the kanaryexporter using the job label in the metric.


kubasikus (approved):
/lgtm

PR #597, opened 2025-06-09
chore: change kanary instance label to tested_cluster

FaisalAl-Rayes:
This is due to the "instance" label being reserved for the IP:PORT of the endpoint the metric is coming from, which was leading our label "instance" to become "exported_instance". This could cause confusion mapping and connecting the metric labels with the code.


kubasikus (approved):
/lgtm
/approve

PR #603, opened 2025-06-11
feat(SPRE-1521): PodNotReady MPC namespace removal

rubenmgx:
There is a huge amount of alerts reaching out to #konflux-misc-channel in relation to PodNotReady for MPC namespace, these are related with Provisioning/cleanup/update pods are not a real indicator that something is happening on MPC namespace (discussed with Infra Team).

We have decided to remove MPC namespace to reduce the noise, since the pods that we really want to monitor are already under konflux_up signal and there is an alert already for that.
@jdcasey @raks-tt

### DIFF
--- a/rhobs/alerting/data_plane/prometheus.pod_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.pod_alerts.yaml
@@ -10,7 +10,7 @@ spec:
     interval: 1m
     rules:
     - alert: UnschedulablePods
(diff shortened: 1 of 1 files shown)


mike-kingsbury (approved):
LGTM

pacho-rh (approved):
LGTM

raks-tt:
Thanks for adding this. LGTM

PR #608, opened 2025-06-12
fix(PVO11Y-4764): kanary alerts should fire after 1s

FaisalAl-Rayes:
This is due to the fact that validation of the alert seems to require a value > 0 for the "for" rule spec.


mftb (approved):
/approve
/lgtm
/ok-to-test

PR #636, opened 2025-06-23
add initial alerting rules for kyverno

glevi-rh:
This PR adds the KyvernoDeploymentDown alert, designed to trigger if Kyverno deployments in the konflux-kyverno namespace are not fully operational.

### DIFF
--- a/rhobs/alerting/data_plane/prometheus.kyverno_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.kyverno_alerts.yaml
@@ -20,7 +20,7 @@ spec:
         annotations:
           summary: Kyverno has reported that not all the deployments pods are up and running for the last 5 minutes.
           description: "The Kyverno Deployment '{{ $labels.deployment }}' in namespace '{{ $labels.namespace }}' on cluster '{{ $labels.source_cluster }}' is unhealthy."
-          alert_routing_key: infra
+          alert_team_handle: <!subteam^S05Q1P4Q2TG>
           team: konflux-infra
           # kyverno/README.md currently in progress.
           runbook_url: https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/main/kyverno/README.md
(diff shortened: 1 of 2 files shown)


rhobs/alerting/data_plane/prometheus.kyverno_alerts.yaml
```diff
@@ -0,0 +1,23 @@
...
+      - alert: KyvernoDeploymentDown
+        expr: |
+          kube_deployment_spec_replicas{namespace="konflux-kyverno"} * 3 != kube_deployment_status_replicas * 3
+        for: 5m
+        labels:
+          severity: warning
+        annotations:
+          summary: kyverno has reported that not all the deployments pods are up and running 5 minutes
+          description: "kyverno on cluster '{{ $labels.source_cluster }}' has that not all the deployments pods are up and running for 5 minutes"
+          alert_team_handle: <!subteam^S05Q1P4Q2TG>
```
TominoFTW, line 22: Since this is non-SLO alert, it needs to have `alert_routing_key` annotation instead of `alert_team_handle`.

Those are getting routed via this [handle](https://gitlab.cee.redhat.com/service/app-interface/-/blame/master/resources/rhobs/production/alertmanager-routes-mst.secret.yaml#L53) for which i believe your team has entry already...
```
                {{ else if or (eq $ark "infra") (eq $ark "admin-checker") (eq $ark "openshift-adp") (eq $ark "cluster-secret-store-rh") (eq $ark "external-secrets-operator") ($ark | match "^multi-platform-controller.*") ($ark | match "^nvme-storage-config.*") -}}
                  <!subteam^S05Q1P4Q2TG>
 ```
If you need to add something to distinguish it like `keyverno` let's update it here so I can create match for it.

Comment (sadlerap) [reply rhobs/alerting/data_plane/prometheus.kyverno_alerts.yaml:22]: AIUI this is the slack handle, so it needs to be the group ID.

File: rhobs/alerting/data_plane/prometheus.kyverno_alerts.yaml [hunk 2]
```diff
@@ -0,0 +1,23 @@
...
+  groups:
+    - name: kyverno-unavailable
+      interval: 1m
+      rules:
+      - alert: KyvernoDeploymentDown
+        expr: |
+          kube_deployment_spec_replicas{namespace="konflux-kyverno"} * 3 != kube_deployment_status_replicas * 3
+        for: 5m
+        labels:
+          severity: warning
```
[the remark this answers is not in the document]
mftb: In that case, this should be converted into a SLO alert.

TominoFTW (approved):
/lgtm
/approve

PR #637, opened 2025-06-23
fix(stoneintg-1225): fixes for original file

jencull:
changed alert to fire after 5 mins instead of 99% uptime edited alert tests to reflect new measurement


TominoFTW (approved):
/lgtm
/approve

PR #638, opened 2025-06-23
chore(SRVKP-7879): Add konflux_up prometheus signals for each Openshift Pipelines component

aThorp96:
Add a `konflux_up` signal based on the number of available replicas for each of the Openshift Pipelines components required by Konflux:
- pipeline-metrics-exporter
- pipeline-metrics-exporter
- pipelines-as-code-controller
- pipelines-as-code-watcher
- pipelines-as-code-webhook
- tekton-chains-controller
- tekton-events-controller
- tekton-operator-proxy-webhook
- tekton-pipelines-controller
- tekton-pipelines-remote-resolvers
- tekton-pipelines-webhook
- tekton-results-api
- tekton-results-api-for-watcher
- tekton-results-retention-policy-agent
- tekton-results-watcher
- tkn-cli-serve

Note that the Tekton Results components are in a different namespace (`tekton-results`) and the components `tekton-pipelines-controller` and `tekton-pipelines-remote-resolvers` are StatefulSets instead of Deployments. There was no prior art for stateful sets in this directory, but I believe I've used equivalent metrics based on the [available metrics](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/metrics/workload/statefulset-metrics.md). 

Resolves https://issues.redhat.com/browse/SRVKP-7879

### DIFF
--- a/rhobs/recording/pipelines_availability_recording_rules.yaml
+++ b/rhobs/recording/pipelines_availability_recording_rules.yaml
@@ -1,0 +1,41 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: rhtap-pipelines-availability
+  labels:
+    tenant: rhtap
+
+spec:
+  groups:
+    - name: pipelines_availability
+      interval: 1m
+      rules:
+        # Deployments
+        - record: konflux_up
+          expr: |
+            clamp_max(
+              label_replace(
+                label_replace(
+                  floor(
+                    kube_deployment_status_replicas_available{
+                      namespace=~"openshift-pipelines|tekton-results"
+                    } / clamp_min(kube_deployment_spec_replicas{namespace=~"openshift-pipelines|tekton-results"}, 1)
+                  ), "check", "replicas-available", "",""
+                ), "service", "$1", "deployment","(.*)"
+              ), 1
+            )
+
+        # Stateful Sets
+        - record: konflux_up
+          expr: |
+            clamp_max(
+              label_replace(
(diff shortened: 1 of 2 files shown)


mftb:
Please add unit tests to your recording rules. You can check several examples in https://github.com/redhat-appstudio/o11y/tree/main/test/promql/tests/recording

rhobs/recording/pipelines_availability_recording_rules.yaml
```diff
@@ -0,0 +1,41 @@
...
+              label_replace(
+                label_replace(
+                  floor(
+                    kube_statefulset_status_replicas_available{
+                      namespace=~"openshift-pipelines|tekton-results"
+                    } / clamp_min(kube_statefulset_replicas{namespace=~"openshift-pipelines|tekton-results"},1)
+                  ), "check", "replicas-available", "",""
+                ), "service", "$1", "statefulset","(.*)"
+              ), 1
+            )
```
aThorp96, line 41: These metrics capture every deployment and statefulset, where the specific services which need to be alerted-on will be defined in the alerts. Each of the respective services are in the unit tests. 

Also note that in the event a service's deployment is scaled to 0 the `clamp_min` will ensure the alert fires.

jdcasey (approved):
This seems like a reasonable set of rules. I like the use of clamp_min, because this can catch sysadmin/sre human error. I suppose the only place we could get in trouble here is if you had a deployment and stateful set with the same name, which is unlikely.

mftb (approved):
/lgtm
/approve

PR #640, opened 2025-06-24
chore(KFLUXVNGD-334): smee konflux_up metric and alerting rule

yftacherzog:
This change adds a recording rule to capture the konflux_up metric for both smee server and client. It then uses that metric for an alerting rule.

The metric is based on a new liveness probe that triggers a live test for the smee client/server for verifying that smee is still serving properly.

See: https://github.com/konflux-ci/smee-sidecar

### DIFF
--- a/rhobs/alerting/data_plane/prometheus.smee_up_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.smee_up_alerts.yaml
@@ -1,0 +1,30 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: rhtap-smee-availability-alerting-rules
+  labels:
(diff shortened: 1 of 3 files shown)


rhobs/alerting/data_plane/prometheus.smee_up_alerts.yaml
```diff
@@ -0,0 +1,30 @@
...
+  groups:
+  - name: smee-availability
+    interval: 1m
+    rules:
+    - alert: SmeeDown
+      expr: |
+        konflux_up{
+          namespace=~"smee|smee-client",
+          check="replicas-available",
+        } != 1
```
FaisalAl-Rayes, line 17: Any reason to why `!= 1` instead of `>= 1`

PR #642, opened 2025-06-24
fix: adjust label_replace for kubearchive

jdcasey:
It appears that the service name may have been set inconsistently compared to other rules (though I'm not convinced this would matter, as it appears to be ironing out the differences in kubearchive deployment names into a single 'kubearchive' from what I can see). I'm also removing the `floor(avg())` combination in favor of just `floor()` in the expression, since the average should be handled at a higher level (I think).

Finally, I wrote a basic test file for it which is derived from the namespace-lister test.


mftb (approved):
/lgtm
/approve

rh-hemartin (approved):

PR #647, opened 2025-06-26
fix(SPRE-1542): Adding recording rule test and fix rule

jdcasey:
The recording rule had a threshold similar to the alert expression, and without a unit test it wasn't detected. This adds the test then fixes the rule.


TominoFTW (approved):
/lgmt
/approve

mftb (approved):
/lgtm
/approve

PR #651, opened 2025-06-30
fix(PVO11Y-4829): Fix %s in availability dashboard

pacho-rh:
The percentages in the availability dashboard were always reporting 100% despite there being known issues with the Konflux service(s). These updated queries should reflect the availability percentages more accurately.


mftb (approved):
/lgtm
/approve

PR #655, opened 2025-07-01
chore(PVO11Y-4833): Enable Kanary alerts to include the type

FaisalAl-Rayes:
This change is done to enable alerting for not just the container kanary signal (and error) but also for RPMs and any other types that would be supported down the road.

Assisted-by: Cursor


kubasikus (changes requested):
almost lgtm, just needs a little readability improvement.

kubasikus (approved):
/lgtm
/approve

PR #664, opened 2025-07-04
Update Konflux UI konflux_up signal

FernandesMF:
This commit lowers the number of replicas that are expected to be up from 3 to 2.

This is because Konflux UI should still be able to function with 2 replicas of each pod, and we don't need to demand that all 3 are up at all times.

This commit changes:
- the konflux_up recording rule
- the tests for the rules


pacho-rh (approved):
LGTM

PR #673, opened 2025-07-07
feat(/SPRE-1597): CPU Usage refinement

rubenmgx:
As of today, the alert was triggered when each individual instance per cluster was failing.
This made the alert to be quite noisy and a single full instance is not a real indicator of an issue.
We have set the alert to fire when there is +2 instances per cluster full which still not 100% an indicator of an issue but that should reduce the amount of alerts we received until we define a proper SLO alert+SOP


martysp21 (approved):
/lgtm

mftb (approved):
/lgtm
/approve
/ok-to-test

PR #677, opened 2025-07-10
fix(SPRE-1623) Api server SOP + http error refinement"

rubenmgx:
Based on https://issues.redhat.com/browse/SPRE-1623 we have added a SOP to handle 5XX error code.
https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/main/infra/sre/api_server_error_rate_alert.md

As well as this, the error rate was triggered based on 429/5XX error codes and since this was causing some false positives, we have removed 429, for which we will just react to 5XX error code.


pacho-rh (approved):
LGTM

mftb (approved):
/lgtm
/approve

PR #690, opened 2025-07-14
MintMaker cluster ram metrics

FernandesMF:
This commit introduces two observatility metrics that will be used in MintMaker's dashboard, and their tests.

The two metrics are related to cluster memory:
- cluster memory requested percentage (as in requested/available)
- percentage of nodes in MemoryPressured state

This is because, if the cluster is experiencing high memory demand, it will probably have issues when processing the MintMaker pipelineruns, and the service might degrade.


pacho-rh (approved):
LGTM

PR #691, opened 2025-07-14
Re-enable MintMaker alerts from 'kflux-rhel-p01'

FernandesMF:
This commit reenables the konflux_up alerts for MintMaker coming from that cluster.

It was previously being set up and unstable, and it was flooding the alerts channel. Now the cluster is stable, and we need to reenable the alerts.


pacho-rh (approved):
LGTM

PR #708, opened 2025-07-24
feat: Enable kanary dashboard to filter by artifact type

FaisalAl-Rayes:
- Added a new variable in the dashboard 'kanary_type'
- Only 1 type could be viewed at a time (this is to avoid complexities in the aggrigation process of same cluster names but with different signals according to the artifact type)

Issue: [PVO11Y-4834](https://issues.redhat.com/browse/PVO11Y-4834)


mftb (approved):
/lgtm
/approve

mike-kingsbury (approved):

PR #711, opened 2025-07-29
chore: Kanary dashboard intial and default values

FaisalAl-Rayes:
This will make it so that the dashboard will by default have all clusters, the production datasource and the container kanary type selected.

Issue: [PVO11Y-4834](https://issues.redhat.com/browse/PVO11Y-4834)


martysp21 (approved):
/lgtm

PR #719, opened 2025-08-03
Added MPC Grafana dashboard

glevi-rh:
## What
Added a new Grafana dashboard for the Multi-Platform Controller (MPC).

## Why
This dashboard improves observability of the MPC by providing key metrics for monitoring its health and performance.


dashboards/grafana-dashboard-MPC.configmap.yaml
```diff
@@ -0,0 +1,2517 @@
...
+          },
+          "pluginVersion": "11.6.3",
+          "targets": [
+            {
+              "datasource": {
+                "type": "prometheus",
+                "uid": "P22466E8E7855F1E0"
+              },
+              "editorMode": "code",
+              "expr": "sum by(platform) (multi_platform_controller_running_tasks{source_cluster=~\"$cluster\", platform=~\"$platform\"})",
```
TominoFTW, line 435: Not sure why those are exported as negative values, maybe it's making sense and I will stand corrected, but just felt like throwing it inhere for clarification - why is that if you happen to know and if it's supposed to be like that, feel free to ignore :) 
```suggestion
              "expr": "abs(sum by(platform) (multi_platform_controller_running_tasks{source_cluster=~\"$cluster\", platform=~\"$platform\"}))",
```
If you happen to change that, don't forget to update the thresholds for colors

TominoFTW (commented):
Preemptive review:

- Rebase the source branch, pipeline should be alright
- One thought about absolute values (maybe if you're not aware of scalar functions in grafana 😄)

(nitpicks but they are setting some format to the repo that keeps things neat) 
- Move the other fields to the top so they are together
- Rename the configmap and the configmap field to reflect that

Other that that lgtm!

dashboards/grafana-dashboard-MPC.configmap.yaml
glevi-rh: It's an open bug so I want to leave it as it is, that metric isn't working well right now, that is why it's also in the "Invalid Metrics" section :)

mshaposhnik:
Is there an URL to staging grafana maybe to get an idea how the dashboard currently looks?

TominoFTW (approved):
lgtm

TominoFTW:
> Is there an URL to staging grafana maybe to get an idea how the dashboard currently looks?

There would be only one created by Gal from his development, if he doesn't have one, this will get deployed to stage and then there will be one under [https://grafana.stage.devshift.net/d/aethy6secwq2oe](https://grafana.stage.devshift.net/d/aethy6secwq2oe).

mshaposhnik (approved):

PR #720, opened 2025-08-04
Fix MintMaker quality recording tests

FernandesMF:
This commit fixes a label that was incorrect in two MintMaker recording tests that was causing them to fail.


TominoFTW (approved):
lgtm

PR #737, opened 2025-08-13
kueue: Make SLOs cluster aware

gbenhaim:
- The alerts weren't aware to the source_cluster label, fixing this.

- Remove the alerts about missing metrics, because for those we can't report which cluster didn't provide the metrics. This kind of alert is more suitable to be included in the Konflux clusters directly.


FaisalAl-Rayes (approved):
/lgtm
/approve

PR #740, opened 2025-08-14
KFLUXINFRA-2089: Kueue ignore code 400 in webhook metric

gbenhaim:
Status 400 is returned when a malformed request is sent to the webhook. This can happen because of user configuration and because of that it's out of scope of this alert (which is about queuing).


sadlerap (approved):

pacho-rh (approved):
LGTM

TominoFTW (approved):
lgtm

PR #752, opened 2025-08-20
fix: Move Kanary alerts to non-slo

jdcasey:
These alerts are not actionable by an SRE, and they flap a lot. Without a LOT more SOP detail (and hopefully, a much finer-grained set of signals and/or tools to help us) this just amounts to noise on our PagerDuty alerts that leads us toward alert fatigue.


jdcasey:
Yeah, I'd agree there...it seems like the validation on these tests is a bit too specific.

mike-kingsbury (approved):
LGTM

ci-operator (approved):
Good call keeping the prior values present & commented.
lgtm

PR #759, opened 2025-08-26
feat(PVO11Y-4899/PVO11Y-4900): Add number of WebRCA incidents and firing alerts

TominoFTW:
Add panels indicating number of ongoing WebRCA incidents and firing SLO alerts for easier readability. With larger timeframe, those are indicating number of incidents/alerts over the period of time.

Updated [dashboard](https://grafana.stage.devshift.net/d/cew71jpnubz0gb/tbehal-test-page?orgId=1&from=now-6h&to=now&timezone=UTC&var-cluster=stone-prod-p02)

Posible scenarios:
image
30 Days, stone-prod-p01:
image

### DIFF
--- a/dashboards/grafana-dashboard-konflux-status-page.configmap.yaml
+++ b/dashboards/grafana-dashboard-konflux-status-page.configmap.yaml
@@ -85,6 +85,19 @@ data:
           "type": "text"
         },
         {
+          "collapsed": false,
+          "gridPos": {
+            "h": 1,
+            "w": 24,
+            "x": 0,
+            "y": 3
+          },
+          "id": 16,
+          "panels": [],
+          "title": "Kanary signals",
+          "type": "row"
+        },
+        {
           "datasource": {
             "type": "prometheus",
             "uid": "P22466E8E7855F1E0"
@@ -151,7 +164,7 @@ data:
                       {
                         "targetBlank": true,
                         "title": "Kanary signal dashboard",
-                        "url": "https://grafana.app-sre.devshift.net/d/eeoimpx3yutq9f/kanary-signal-dashboard?var-kanary_clusters=${cluster}"
+                        "url": "https://grafana.app-sre.devshift.net/d/eeoimpx3yutq9f/kanary-signal-dashboard?var-kanary_clusters=${cluster}&var-kanary_type=container"
                       }
                     ]
                   }
@@ -163,7 +176,7 @@ data:
             "h": 6,
             "w": 12,
             "x": 0,
-            "y": 3
+            "y": 4
           },
           "id": 10,
           "options": {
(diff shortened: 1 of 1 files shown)


dashboards/grafana-dashboard-konflux-status-page.configmap.yaml
```diff
@@ -151,7 +164,7 @@ data:
                       {
                         "targetBlank": true,
                         "title": "Kanary signal dashboard",
-                        "url": "https://grafana.app-sre.devshift.net/d/eeoimpx3yutq9f/kanary-signal-dashboard?var-kanary_clusters=${cluster}"
+                        "url": "https://grafana.app-sre.devshift.net/d/eeoimpx3yutq9f/kanary-signal-dashboard?var-kanary_clusters=${cluster}&var-kanary_type=container"
```
pacho-rh, line 167: The `kanary_type` looks to be some leftover variable that can be removed.

dashboards/grafana-dashboard-konflux-status-page.configmap.yaml
```diff
@@ -191,51 +204,302 @@ data:
...
+                  "options": "Value"
+                },
+                "properties": [
+                  {
+                    "id": "links",
+                    "value": [
+                      {
+                        "targetBlank": true,
+                        "title": "Kanary signal dashboard",
+                        "url": "https://grafana.app-sre.devshift.net/d/eeoimpx3yutq9f/kanary-signal-dashboard?var-kanary_clusters=${cluster}&var-kanary_type=rpm"
```
pacho-rh, line 283: The `kanary_type` looks to be some leftover variable that can be removed.

pacho-rh (changes requested):

dashboards/grafana-dashboard-konflux-status-page.configmap.yaml
pacho-rh: Ah my mistake

pacho-rh:
> The alert list (talking about the specific predefined panel) cannot show historical data at all.

Yes unfortunately that is the case with this panel type. I think we discussed this before. We'll have to use a different panel type with the `ALERTS` metrics to get our historical alerts. We can handle that separately. I'll be looking into it.

> So it will always show data for set timeframe. I can look if there is possibility to show only some relevant timeframe (hardcode it), lets say 10 minutes, so it is not showing number of alerts from the past 6hours for example 🤷

If anything you could use something like `last_over_time`. Does not solve the entire problem. For instance, if the user selects some specific timeframe instead of the last x time, the value shown will be for last given value in that timeframe and not what is currently happening.
See if you can figure out that hardcoded timeframe. And do reach out to me if you would like help with it and we can look at it together.

TominoFTW:
Updated panels look like this and are displaying only the last 10 minutes. Grafana timepicker serves only as selector for WebRCA table as of now. 
image
image

PR #761, opened 2025-08-28
feat(SPRE-1837) Adding alerts grafana dashboard

rubenmgx:
We have created a dashboard to monitor SLO/non-SLO Alerts, so it makes easier for us to look for alets that are firing quite often and needs to be refined.


TominoFTW (approved):
lgtm

PR #762, opened 2025-08-29
feat(SPRE-1875): Include cluster in release service alerts

gcpsoares:
- Include cluster in release service alerts.


pacho-rh (approved):
LGTM

PR #773, opened 2025-09-04
fix(stoneintg-1321): tidy up y axis info on 4 panels

jencull:


jencull:
@mftb for your viewing pleasure :)

PR #781, opened 2025-09-09
feat(STONEINTG-1323): add time to start PLR panel

jsztuka:
Adding **time to start PLR panel** for integration service SLO.


FaisalAl-Rayes (approved):
/lgtm
/approve

PR #782, opened 2025-09-10
Stoneintg 1324

jencull:
Integration Service has an SLO for latency release creation. This was previously
deactivated. Reactivating and adding new runbook url.


jencull:
closed due to incorrect ticket reference

PR #796, opened 2025-09-17
feat(SPRE-1880) API Server alert

rubenmgx:
Based on the ongoing work on SLO Phase II, we are implementing couple of alerts for API Server:

- KonfluxAPIServerHighCPUUsage - CPU Usage for API Server Job (per cluster) above 80%
- KonfluxAPIServerHighMemoryUsage - Memory Usage for API Server Job (per cluster) above 80%

Thresholds are not 100 Accurate right now since we dont have previous examples, but the choose % should be a real indicator. There will be some refinement for some components done during Q4, so there will be a follow up on this alert in case we need to make some changes.

Related SOPs are in https://gitlab.cee.redhat.com/konflux/docs/sop/-/tree/main/infra/sre


raks-tt:
Thank you. this looks good.

FaisalAl-Rayes (approved):
/lgtm
/approve

PR #802, opened 2025-09-19
MintMaker DUC alerting

FernandesMF:
This commit implements an alert in case the DUC creation is not taking place as expected.

DUCs are a custom resource in MintMaker that acts as triggers. If they are not created at the expected times, we should fire a warning.

Assisted by: Cursor


martysp21 (approved):
/lgtm

PR #991, opened 2026-02-01
Add workload_priority filter to kueue admission wait time alert

gbenhaim:
Filter out konflux-dependency-update priority class from admission wait time calculations. This class is used by the mintmaker pipeline where it's expected to have high admission wait times.

Assisted-By: Cursor


kubasikus:
@gbenhaim does this still require our attention?

raks-tt:
LGTM, @gbenhaim - please resolve the conflicts.

PR #1077, opened 2026-04-15
fix(PVO11Y-5178): Set imagePullPolicy to Always for kanary exporter container

FaisalAl-Rayes:
The default IfNotPresent policy was preventing updated images from being pulled, causing stale versions to run.

Issue: PVO11Y-5178


TominoFTW (approved):

FaisalAl-Rayes:
Closed as there was an image reference that would be changed in infra-deployments

PR #1262, opened 2026-08-05
fix(SPRE-5243:RegistryExporterNodeFailure)change alert to fire on >50…

sjamode-rh:
### Description

- Changes RegistryExporterNodeFailure from a per-node alert to a cluster-level ratio alert.
- The for: 15m only guarantees the >50% ratio held for 15 minutes, not
  per-node continuous failure
- Updates unit tests to match new aggregated semantics.

## Why
Single-node failures during rolling node replacements were generating noise. The new threshold only alerts when a significant portion of nodes are affected.

## Testing
Promtool unit tests pass for all scenarios.

---


pacho-rh (approved):
LGTM

raks-tt (approved):
/lgtm

PR #1266, opened 2026-08-06
fix(alert-tuning): [SPRE-5878] suppress EtcdHighNumberOfLeaderChanges during cluster upgrades

ay-garg:
## Summary

Resolves [SPRE-5878](https://redhat.atlassian.net/browse/SPRE-5878) — Alert Review & Tuning: `EtcdHighNumberOfLeaderChanges`

Suppresses the alert per-cluster during active OCP upgrades using a recording
rule as the inhibition signal, while preserving full sensitivity (`> 2`)
during normal operation.

---

`EtcdHighNumberOfLeaderChanges` generated **~109 false positive firings** across
two consecutive 14-day investigation windows with **0% actionable signal** and
zero associated PagerDuty incidents or user-impacting events.

Every firing occurred strictly within scheduled weekly OCP upgrade windows.
During a rolling upgrade, the MachineConfigOperator drains and reboots control
plane nodes one at a time — each restart forces a Raft leader re-election. A
standard 3-node etcd cluster produces 2–3 elections per upgrade (~70-minute
window), mathematically guaranteeing a crossing of the `> 2` threshold on every
upgrade cycle.

---

## Why not raise the threshold

Raising the threshold from `> 2` to `> 4` was considered and rejected. The
maximum observed `increase()` value during any upgrade was **3.158** — raising
to `> 4` would eliminate all false positives, but creates a blind spot: genuine
instability (e.g. disk pressure, network blip) causing 3–4 leader elections
during normal operation would go silently undetected. Preserving full sensitivity
(`> 2`) is the correct behaviour.

---

## Resolution

Suppress the alert **per-cluster** during active upgrades using the Prometheus
`unless` operator with a dedicated recording rule as the inhibition signal:

```promql
(increase(etcd_server_leader_changes_seen_total{namespace="openshift-etcd"}[10m]) > 2)
unless on (source_cluster)
cluster_upgrade_ongoing

Why a recording rule instead of ALERTS{} directly

The unless clause references cluster_upgrade_ongoing (a recording rule)
rather than ALERTS{alertname="ClusterUpgradeOngoing"} directly, for two
reasons:

1. alertstate ambiguity — ALERTS{} silently matches both
alertstate="firing" and alertstate="pending". The recording rule
explicitly filters alertstate="firing" — suppression only activates once
the upgrade is confirmed, not while it is still within its for: 1m window.
2. Stability — ALERTS{} is an internal Prometheus construct. A named
recording rule produces a stable, well-defined, reusable time series that
other rules can consume in the future.

Label alignment

cluster_name on aus_cluster_upgrade_policy_info maps 1:1 to
source_cluster on etcd_server_leader_changes_seen_total across all RHTAP
clusters (verified in production RHOBS — short-name format, e.g.
kflux-prd-rh03). A label_replace() on the ClusterUpgradeOngoing alert
expression copies cluster_name → source_cluster, enabling the per-cluster
join.

---
Files changed

┌─────────────────────────────────────────────────────────────────────────┬──────────────────────────────────────────────────────┐
│                                  File                                   │                        Change                        │
├─────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────┤
│                                                                         │ New — recording rule that materialises               │
│ rhobs/recording/cluster_upgrade_recording_rules.yaml                    │ cluster_upgrade_ongoing{source_cluster}=1 only when  │
│                                                                         │ ClusterUpgradeOngoing is alertstate="firing"         │
├─────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────┤
│                                                                         │ Wrapped expression in label_replace() to expose      │
│ rhobs/alerting/data_plane/prometheus.cluster_ongoing_updrade_alert.yaml │ source_cluster from cluster_name, enabling           │
│                                                                         │ per-cluster join via unless on (source_cluster)      │
├─────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────┤
│                                                                         │ Added unless on (source_cluster)                     │
│ rhobs/alerting/data_plane/prometheus.cluster_capacity_alerts.yaml       │ cluster_upgrade_ongoing inhibition; threshold        │
│                                                                         │ remains > 2                                          │
├─────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────┤
│ test/promql/tests/data_plane/cluster_upgrade_ongoing_alert_test.yaml    │ Added source_cluster label assertion (now present    │
│                                                                         │ due to label_replace)                                │
├─────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────┤
│                                                                         │ Fixed pre-existing vacuous test (missing namespace   │
│ test/promql/tests/data_plane/cluster_capacity_test.yaml                 │ label); added suppression, per-cluster specificity,  │
│                                                                         │ and stable-counter test cases                        │
├─────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────┤
│                                                                         │ New — unit tests for recording rule:                 │
│ test/promql/tests/recording/cluster_upgrade_recording_rules_test.yaml   │ alertstate="firing" → signal=1; alertstate="pending" │
│                                                                         │  → no output                                         │
└─────────────────────────────────────────────────────────────────────────┴──────────────────────────────────────────────────────┘

---
Test results

promtool test rules:
  cluster_capacity_test.yaml                  SUCCESS
  cluster_upgrade_ongoing_alert_test.yaml     SUCCESS
  cluster_upgrade_recording_rules_test.yaml   SUCCESS

Validated scenarios:
- Leader changes > 2, no upgrade in progress → alert fires ✅
- Leader changes > 2, upgrade ongoing on same cluster → suppressed ✅
- Leader changes > 2, upgrade ongoing on different cluster → alert fires ✅
- Counter stable (no leader changes) → no alert ✅

---

[SPRE-5878]: https://redhat.atlassian.net/browse/SPRE-5878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rhobs/alerting/data_plane/prometheus.cluster_capacity_alerts.yaml
+++ b/rhobs/alerting/data_plane/prometheus.cluster_capacity_alerts.yaml
@@ -79,7 +79,9 @@ spec:
 
       - alert: EtcdHighNumberOfLeaderChanges
         expr: |
-          increase(etcd_server_leader_changes_seen_total{namespace="openshift-etcd"}[10m]) > 2
+          (increase(etcd_server_leader_changes_seen_total{namespace="openshift-etcd"}[10m]) > 2)
+          unless on (source_cluster)
+          cluster_upgrade_ongoing
         for: 1m
         labels:
           severity: warning
--- a/test/promql/tests/data_plane/cluster_capacity_test.yaml
+++ b/test/promql/tests/data_plane/cluster_capacity_test.yaml
@@ -140,7 +140,7 @@ tests:
 
   - interval: 1m
     input_series:
-      # Etcd No Leader failures within threshold limit, so it will be alerted.
+      # Leader changes exceed threshold, no upgrade in progress — alert fires.
       - series: 'etcd_server_leader_changes_seen_total{namespace="openshift-etcd",source_cluster="cluster01"}'
         values: '3+2x60'
 
@@ -162,8 +162,46 @@ tests:
 
   - interval: 1m
     input_series:
-      # Etcd has Leader within threshold limit, so it will not be alerted.
-      - series: 'etcd_server_leader_changes_seen_total{source_cluster="cluster01"}'
+      # Leader changes exceed threshold BUT a cluster upgrade is ongoing for the same cluster.
+      # The unless operator suppresses the alert — upgrade-induced re-elections are expected behaviour.
+      - series: 'etcd_server_leader_changes_seen_total{namespace="openshift-etcd",source_cluster="cluster01"}'
+        values: '3+2x60'
+      - series: 'cluster_upgrade_ongoing{source_cluster="cluster01"}'
+        values: '1+0x60'
+
+    alert_rule_test:
+      - eval_time: 11m
+        alertname: EtcdHighNumberOfLeaderChanges
+
+  - interval: 1m
+    input_series:
+      # Leader changes exceed threshold on cluster01, but the upgrade is ongoing on a DIFFERENT cluster (cluster02).
+      # Suppression is per-cluster — cluster01 alert must still fire.
+      - series: 'etcd_server_leader_changes_seen_total{namespace="openshift-etcd",source_cluster="cluster01"}'
+        values: '3+2x60'
+      - series: 'cluster_upgrade_ongoing{source_cluster="cluster02"}'
+        values: '1+0x60'
+
+    alert_rule_test:
+      - eval_time: 11m
+        alertname: EtcdHighNumberOfLeaderChanges
+        exp_alerts:
+          - exp_labels:
+              component: etcd
+              namespace: openshift-etcd
+              severity: warning
+              source_cluster: cluster01
+            exp_annotations:
+              summary: 'Etcd high number of leader changes'
+              description: 'Etcd leader changed more than 2 times during 10 minutes in cluster cluster01.'
+              alert_routing_key: spreandinfra
+              team: o11y
+              slo: 'false'
+
+  - interval: 1m
+    input_series:
+      # Counter is stable — no leader changes, no alert.
+      - series: 'etcd_server_leader_changes_seen_total{namespace="openshift-etcd",source_cluster="cluster01"}'
         values: '1+0x60'
 
     alert_rule_test:
--- a/rhobs/alerting/data_plane/prometheus.cluster_ongoing_updrade_alert.yaml
+++ b/rhobs/alerting/data_plane/prometheus.cluster_ongoing_updrade_alert.yaml
@@ -12,9 +12,12 @@ spec:
       rules:
         - alert: ClusterUpgradeOngoing
           expr: >-
-            (aus_cluster_version_remaining_soak_days <= -2 >= -3)
-            * on(cluster_uuid) group_left(cluster_name, current_version)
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

File: test/promql/tests/data_plane/cluster_capacity_test.yaml [hunk 1]
```diff
@@ -155,14 +155,24 @@ tests:
...
+      - series: 'etcd_server_leader_changes_seen_total{namespace="openshift-etcd",source_cluster="cluster01"}'
+        values: '0 0 0 3 3 3 3 3 3 3 3 3'
+
+    alert_rule_test:
+      - eval_time: 11m
+        alertname: EtcdHighNumberOfLeaderChanges
+
+  - interval: 1m
+    input_series:
+      # Counter is stable — no leader changes, no alert.
```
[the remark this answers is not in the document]
ay-garg: Fixed in the latest commit — added namespace="openshift-etcd" to the stable counter test series so it now correctly matches the alert expression and genuinely validates the no-alert scenario.

rhobs/alerting/data_plane/prometheus.cluster_capacity_alerts.yaml
```diff
@@ -79,7 +79,7 @@ spec:
 
       - alert: EtcdHighNumberOfLeaderChanges
         expr: |
-          increase(etcd_server_leader_changes_seen_total{namespace="openshift-etcd"}[10m]) > 2
+          increase(etcd_server_leader_changes_seen_total{namespace="openshift-etcd"}[10m]) > 4
```
gcpsoares: If the false alerts only happen during upgrades, another way would be change the promql expression to not trigger the leader change alert during the ongoing upgrades. 
[Upgrade Ongoing Alert](https://github.com/redhat-appstudio/o11y/blob/main/rhobs/alerting/data_plane/prometheus.cluster_ongoing_updrade_alert.yaml#L13)

gcpsoares (commented):
Check if we can inhibit the alert during ongoing upgrades.

rhobs/alerting/data_plane/prometheus.cluster_capacity_alerts.yaml
ay-garg: You're absolutely right @gcpsoares — after thinking through the trade-offs more carefully, the threshold approach does introduce a blind spot between 2 and 4 that could silently miss genuine instability (e.g., disk pressure or brief network partitions causing 3 leader elections). That's not an acceptable trade-off.

The inhibition approach is the correct fix here — it preserves full sensitivity (`> 2`) during normal operation while suppressing the noise specifically during active upgrade windows, which is exactly what we need.

The main blocker is the label mismatch: `ClusterUpgradeOngoing` carries `cluster_name` while `EtcdHighNumberOfLeaderChanges` uses `source_cluster`, which breaks the per-cluster `unless` join. I'm looking into aligning those labels so the inhibition works correctly at the cluster level.

Will update the PR once I have a clean implementation. Thanks for pushing back on this — it's the right call.

gcpsoares (approved):
/lgtm

TominoFTW (approved):

